PR #5, opened 2018-08-20
Add support for tuple-structs and unit-structs

omni-viral:
subj

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -15,6 +15,21 @@ struct Foo {
     z: i32,
 }
 
+#[derive(PartialEq, TypedBuilder)]
+struct Bar(
+    i32,
+    // #[default] without parameter - use the type's default
+    #[default]
+    Option<i32>,
+
+    // Or you can set the default(encoded as string)
+    #[default="20"]
+    i32,
+);
+
+#[derive(PartialEq, TypedBuilder)]
+struct Baz;
+
 fn main() {
     assert!(
         Foo::builder().x(1).y(2).z(3).build()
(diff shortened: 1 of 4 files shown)


idanarye:
1. Please add a test
2. Why would one even need a builder for a unit type? It doesn't offer completeness because we don't support enums, and it doesn't offer any genericness because we don't have a builder trait.

omni-viral:
1. Ok.
2. It can become complete with added enum support.

idanarye:
I'm actually kind of worried about enum support. Enum support would mean we could do this:

```rust
#[derive(TypedBuilder)]
enum Foo {
    Bar {
        a: i32,
        b: i32,
        c: i32,
    },
    Baz {
        e: i32,
        f: i32,
        g: i32,
    },
}

assert!(Foo::build_bar().a(1).b(2).c(3).build() == Foo::Bar { a: 1, b: 2, c: 3 });
assert!(Foo::build_baz().d(4).e(5).f(6).build() == Foo::Baz { d: 4, e: 5, f: 6 });
```

But... I consider having complex structs as enum variants an anti-pattern, because then you need to destruct it every time you want to access any of the fields. It's better to use nested types:

```rust
#[derive(TypedBuilder)]
struct Bar {
    a: i32,
    b: i32,
    c: i32,
}

#[derive(TypedBuilder)]
struct Baz {
    e: i32,
    f: i32,
    g: i32,
}

#[derive(TypedBuilder)]
enum Foo {
    Bar(Bar),
    Baz(Baz),
}
```

But then you wouldn't need a builder for `Foo`, because you would just do:

```rust
Foo::Bar(Bar::build().a(1).b(2).c(3).build());
Foo::Baz(Baz::build().d(4).e(5).f(6).build());
```

Now, it would be nice if we could have `Foo::build_bar()` and `Foo::build_baz()` for that style as well - but I don't think this is possible with Rust's current type system and macro system.

Issue #19, opened 2019-10-27
Add strip_option attribute


mwilliammyers:
Nice crate! 

It would be super handy to have an option to strip `Option` in addition to being generic over `Into` like [`derive-builder`](https://github.com/colin-kiegel/rust-derive-builder#usage-and-features).

Example:

```rust
#[derive(TypedBuilder)]
struct User {
    id: String,
    #[builder(strip_option)]
    name: Option<String>,
}

User::builder().id("id").name("bob").build();
```

instead of having to do:
```rust
User::builder().id("id").name(Some("bob".to_string())).build();
```

What do you think?

idanarye:
Semantically, this would only make sense for non-mandatory fields. `.name(None)` won't work, because the argument type of `name` is `String` ([assuming we remove the auto-`into()`](https://www.reddit.com/r/rust/comments/eqjktn/community_thoughts_about_introducing_a_breaking/feufnzd/)), so the only way to use the builder to create a nameless user is to make `name` `#[builder(default, strip_option)]`, and not call its setter.

So, I wonder what's the best way to extern this in the API.

* Disallow `strip_option` without a `default`?
* Automatically add a `default`? (maybe pick a name the better represents this?)
* Add a `name_none()` setter? (yea, that's probably a bad idea)

I also had this idea for an aggregating field. Something like:
```rust
#[derive(TypedBuilder)]
struct Foo {
    #[builder(collect)]
    numbers: Vec<i32>,
}

assert_eq!(
    Foo::builder().number(1).number(2).number(3).build(),
    vec![1, 2, 3]);
```
(yea, I probably want to avoid the singular-plural hell...)

This might be related to your idea, because `Option` in Rust (and FP in general) can be treated as a collection of zero or one items.

Then again - with `Option` we want to disable the setter after the first invocation, so maybe this is not such a good idea...

idanarye:
I also don't really like the name `strip_option`. It implies that something is an `Option` and you are converting it to non-`Option`, when in practice it's the other way around (you are accepting something that is not `Option` and converting it to `Option`)

This flag is similar to the auto-into flag (which's name is also yet to be decided), so maybe it's a good idea to give them similar names? Like `#[builder(auto_into)]` and `#[builder(auto_some)]` (since we are automatically slapping `Some()` on the argument)? Or maybe, since `into()` makes `Some()` redundant, I should unite them to `#[builder(auto=into)]` and `#[builder(auto=Some)]`?

[...]

PR #29, opened 2020-01-17
Generic type parameter inference

guoli-lyu:
## Description
Currently, all setters use `Into` trait to gain a flexible and ergonomic API. 

For example, for `Foo{s:String}`, a setter with a signature of `fn s<T: Into<String>>(input:s)` can now accept both `&str` and `String`. 

However, `Into` conversion design confuses rustc when there is any generic type. When the setter get a concrete type, there is no way that rustc can infer what type to convert into. So we have to explicitly describe the generic type, for instance, `Foo::<uszie>::builder().x(0).build()` for `Foo<T>{x:T}`. 

This is really annoying when the number of generic types increases.

## Possible Solution

Maybe we should offer a method to disable `Into` conversion.

A good way is to implement a field attribute, say `skip_into`, so user can disable conversion on demand.

```rust
    #[derive(TypedBuilder)]
    struct Foo<X> {
        #[builder(skip_into)]
        x: X,
    }
```

By default, `Into` conversion is still enabled for all field setters, so there is no need to modify previous code.

Well, I think there must be a better name for `skip_into`, but it just eludes my mind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- `#[builder(no_into)]` attribute on fields, to skip a `Into` conversion from 
+- `#[builder(skip_into)]` attribute on fields, to skip a `Into` conversion from 
   input variable to target type, and thus enable rustc inference for generic type.
 
 ## 0.4.1 - 2020-01-17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- `#[builder(no_into)]` attribute on fields, to skip a `Into` conversion from 
+  input variable to target type, and thus enable rustc inference for generic type.
 
 ## 0.4.1 - 2020-01-17
 ### Fixed
--- a/src/field_info.rs
+++ b/src/field_info.rs
@@ -58,7 +58,7 @@ impl<'a> FieldInfo<'a> {
 pub struct FieldBuilderAttr {
     pub doc: Option<syn::Expr>,
     pub skip: bool,
-    pub no_into: bool,
+    pub skip_into: bool,
     pub default: Option<syn::Expr>,
 }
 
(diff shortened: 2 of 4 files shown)


idanarye:
I'm also pondering the idea of introducing a breaking change and _not_ having `Into` as the default. I think maybe we can guess the defaults according to some heuristics - say do it for strings, don't do it for anything else - and allow `#[builder(into=true)]` or `#[builder(into=false)]` to set this decision manually.

guoli-lyu:
Why disable `Into` by default? 

When the type is known, the `Into` trait does offer a more generic setter. I personally love the `Into` and `From` conversion when exposing a public API.

idanarye:
The implicit `into()` makes sense initially (which is why I did it), but once you start using `TypedBuilder` you encounter some limitations it imposes. For example:

```rust
    #[derive(typed_builder::TypedBuilder)]
    struct Foo {
        bar: usize,
        baz: usize,
    }

    Foo::builder()
        .bar(1usize) // will not compile without the literal suffix
        .baz("2".parse::<usize>().unwrap()) // will not compile with the turbofish
        .build();
```

`.bar(1)` would not work, because Rust by default treats integer literals as signed, and signed literal cannot `into()` unsigned ones because you'd lose the sign. Without the automatic `into()` it would work, because the argument of `bar()` would be a `usize` so Rust will know to make the literal a `usize`.

`.baz("2".parse().unwrap())` would not work, because Rust cannot determine what to parse the string as. It needs to parse it to something, and then convert that something to `usize`, but it has no way to determine the concrete type of that something.

I'm also starting to have second thoughts about doing it for strings. Yes, it's convenient, but it hides heap allocation - and the Rust philosophy is to make expensive operations stand out. So removing it entirely will make this crate more rusty.

I'm not against making it opt-in though. Which will also result in a better named flag.

idanarye:
I'm asking the Rust community in Reddit about this - https://www.reddit.com/r/rust/comments/eqjktn/community_thoughts_about_introducing_a_breaking/

guoli-lyu:
Well, reading through the reddit thread, I am still not sure about the pros and cons on switching off `Into` by default.

However, I appreciate the way to offer two method, a `bar(_)` and `bar_from(_)`. In this case, it's quite explicit and there is no need for a field attribute like #[builder(skip_into)]` or whatever the attribute name. A `bar(_)` only accepts exactly the same type and no hidden allocation while a `bar_from(_)` might have allocation depending on the input.

Though I against the `bar_try_from` method, which violate the goal to keep all validation for builder at compile time. I love the current 'infallible at runtime' builder.

idanarye:
I'm merging this as-is, but I'll change it before releasing a version. I also want to do #19 in the same release, since I find them quite related and want to decide on a uniform naming convention before I commit(ish - it's still pre-1.0.0) to the API.

guoli-lyu:
Thank you for you great work.

idanarye:
Thank you for your contribution.

PR #37, opened 2020-05-15
Add ability to use both StripOption and AutoInto

asonix:
I accidentally rustfmt'd the code and I can clean up this PR tomorrow without the formatting changes, but here's a change to support Into for stripped option types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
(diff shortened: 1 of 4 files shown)


idanarye:
Mmm... for some reason I was under the impression that `.into()` also converts the value inside the option making `strip_option` redundant, but I tested it isn't so. Knowing Rust's (lack of) specialization rules, I don't know why I thought it'll work.

Anyway, other than the cleaning you said you'll do please also add a test and a changelog entry.

asonix:
The PR has been cleaned, and tests have been added

idanarye:
OK. I might change the implementation though - it made sense to have `SetterArgSugar` as an `enum` when these flags were mutually exclusive, but now that they can be used together we may want to represent them as two booleans.

PR #42, opened 2020-11-20
Fix lifetime bounds erroneously preserved in phantom generics

Tamschi:
It was a bit tricky to figure this out (since I'm using this macro nested inside another much larger one), but I think I've gotten better at debugging proc macros with invalid output now 😅

In any case, thanks for making this crate! It's going to save me a huge amount of work.
I've included the test for this issue a commit earlier than the fix, so if you'd like to preserve bisectable history, please squash them when merging.


idanarye:
Thanks!

PR #46, opened 2021-01-17
Optionally transform `Extend` fields to add items one by one

Tamschi:
This is only a proof of concept right now, and has many rough edges. I decided to make a draft PR early to discuss this more easily.

Thanks for your consideration.

- - -

# Motivation

I'm working on a web frontend framework where I use `typed-builder` to enable type-safe named and optional parameters, like this:

```rust
<*Custom .a = {1} .b = {"B"}>
<*Custom .b = {"B"} .a = {1}>
```

Something I'd like to add to this (without special syntax) is a way to declare arguments as accepting zero-or-more or one-or-more parameters:

```rust
<*Custom .a = {1} .a = {2} .a = {3} .b = {"B"}>
<*Custom .b = {"B"} .b = {"C"} .b = {"D"}>
```

I could probably also do this on my end, but the feature is most likely a bit cleaner and easier to implement here in this crate.

# Planned Changes

For now, I added a setter setting `strip_extend` that optionally takes an expression to transform the first given argument into the collection type (which is necessary to make this work with, for example, [`Vec1`](https://lib.rs/crates/vec1)).
I'll have to think about the naming here, maybe use two different names to make it clear that the parametrised variant needs to process the the argument too, instead of just providing an empty seed value.

The unparametrised version uses a specified `default =` or `Default::default()` to seed the field, but a field that's not set at all is still rejected unless `default` is specified explicitly.

The generated setter method itself is generic over `<Argument>`, which means for example `Vec` will accept both its item type and references to it (if the items are `Copy`). This doesn't *seem* to cause too many issues with type inference, but the type can also be specified explicitly when calling the method.
This could cause issues with custom initialisers though, since `Argument` can only be used to call `.extend(iter::once(...))` right now. I'm not yet sure how to fix this.

Other issues with this draft that still need to be resolved:

- [ ] Extending by an iterator

  This probably shouldn't be missing, but would need another method name. (`"extend_{field_name}"`?)

- [ ] Compatibility with `into`/`auto_into`

  This might need a bit of work so that `Argument` can still be specified explicitly on the builder method call.

- [ ] Compatibility with `strip_option`

  I think this can be done in a hacky way without many changes, but it would definitely be cleaner if the field value was only wrapped in `Some` once `.build()` is called. (I'm not sure whether this would affect performance, but more importantly it's a breaking change regarding what `TypedBuilderFields` is set to.)

- [ ] Tests
- [ ] Documentation


Tamschi:
The public API feels right to me now, as far as implemented. The _first_ setter call for a given collection field isn't generic anymore, which simplifies and solves typing problems around custom collection initialisers and avoids any potential issues with type inference.

Writing a custom initialiser isn't entirely intuitive in that it **should** also process the first item (which is necessary for strictly non-empty collections, as mentioned above). For this reason, I added the following error on there:

image

I don't think the error location can be improved nicely with current proc macro tooling, but this could get sorted out with either upcoming changes to `Span`s or a Syn update building on them. The lint level definition is located on the `=` token.

idanarye:
Can you please not run the code through a formatter? If the code needs to be formatted it deserves a PR of its own, where the settings of the formatter can be discussed and more importantly - where there is no need to go through all the code changes because it's auto-generated anyways and contains no real changes. When you do it as part of a regular PR, most of the changes are formatting changes which makes it very hard to review the real changes.

Regarding the PR itself - I've only skimmed it (I can't properly read it with all the auto-generated formatting noise), but a few remarks:

* I'm not sure `strip_collection` is the right term. I think `extend` better conveys what t his feature does (each call to the setter extends the collection)
* I'm not sure how I feel about the custom initizlizer:
  * At the very least, it should be `#[builder(setter(extend(initializer = ...)))]` and not `#[builder(setter(extend = ...))]` because it's not intuitive that the primary argument of `extend` is the custom initializer.
  * Is it even required? Does it have real use-cases? I understand it can be used to start a collection with already existing entries, but every other field type requires the user to either use the default or set the entire content of the field. We don't have a setting for string fields to prefix the set strings with something...
* I don't think we need the `HashMap` in the `example.rs`. It makes it too cumbersome.
* Missing:
  * Documentation (goes in `lib.rs`)
  * Changelog entry.
  * Tests.

Tamschi:
Thanks for the feedback (and thank you for considering this feature submission at all)!

> Can you please not run the code through a formatter? If the code needs to be formatted it deserves a PR of its own, where the settings of the formatter can be discussed and more importantly - where there is no need to go through all the code changes because it's auto-generated anyways and contains no real changes. When you do it as part of a regular PR, most of the changes are formatting changes which makes it very hard to review the real changes.

Sure, and sorry about that. Not using the auto-formatter at all makes it pretty difficult for me personally to work on code, but I'll add a commit that reverts those changes once I take care of the other problems with my changes.

Feel free to unsubscribe from the pull-request until then if you'd like to avoid progress notifications - I'll ping you via @ mention after I clean it up.

> Regarding the PR itself - I've only skimmed it (I can't properly read it with all the auto-generated formatting noise), but a few remarks:
> 
> * I'm not sure `strip_collection` is the right term. I think `extend` better conveys what t his feature does (each call to the setter extends the collection)

It's a bit more complicated, since (as far as I can tell) collections aren't thoroughly defined by a single trait.
Right now I'm using `Extend` for repeat calls, but `IntoIterator::Item` to strictly constrain the argument on the first call and `FromIterator` for the default collection initialiser if no other default is present. This mirrors how I'd hand-write such a builder.

It's definitely possible to reduce this to just `Extend`, but I'm concerned that would hurt ergonomics and could have a slight performance impact.

I'm open to the rename regardless of the above, though.
I originally chose `strip_extend`/`strip_collection` for consistency with `strip_option`, since the argument type is "unwrapped" in a similar way.

> * I'm not sure how I feel about the custom initizlizer:
>   
>   * At the very least, it should be `#[builder(setter(extend(initializer = ...)))]` and not `#[builder(setter(extend = ...))]` because it's not intuitive that the primary argument of `extend` is the custom initializer.

That sounds a lot better than what I came up with, thanks! I'll change it 👍

>   * Is it even required? Does it have real use-cases? I understand it can be used to start a collection with already existing entries, but every other field type requires the user to either use the default or set the entire content of the field. We don't have a setting for string fields to prefix the set strings with something...

It's an edge case, I think. For the default collections in `std` it's not necessary at all.

Collections that by type require at least one entry are more ergonomic in some use-cases, in that you don't have to `.unwrap()` any return values from e.g. `.first()` and `.last()`. This also means they mostly can't implement `Default` and usually won't implement `FromIterator`, though.
There doesn't seem to be a standard way to move a single item into a new collection, which makes configuration necessary.

I'd like to have this feature in my framework because it's, in my experience, quite likely to get GUI designs where the empty case is under-specified. If that's actually not expected to be possible, I'd rather mark it in the type system than add it as comment and try to invent a style that looks sort of okay when the component is misused.

That said, in my particular framework I'll likely be able to do this transparently even without specific support here.
I just think it would be cleaner to add support for it inside the parameter and builder types, instead.

The (existing) `String` case is different to me in that the value is set all at once. A custom `extend(initializer = …)` shouldn't be used to add a prefix to the collection, but I don't see a way to actually prevent this.

It could also be used to start a collection _without_ existing elements if the default has some (since I'd like to have the `extend` setters strictly default to extending the value). That to me seems a bit more likely as a way to have a placeholder, but I personally would much rather use a construct like

```rust
#[builder(default, setter(strip_option, extend(initializer = vec1![entry_first])))]
entry: Option<Vec1<Entry>>,
```

in that case (which would most likely be written as `entry+?: Entry,` in my macro, once I implement it).

> * I don't think we need the `HashMap` in the `example.rs`. It makes it too cumbersome.

I'll remove it. I think it would be a good idea to demo the `extend` feature in a separate example, so I'll revert the changes to this file if you don't have any objections.

> * Missing:
>   
>   * Documentation (goes in `lib.rs`)
>   * Changelog entry.
>   * Tests.

I'll add it to my checklist in PR summary 👍

idanarye:
> Thanks for the feedback (and thank you for considering this feature submission at all)!
> 
> > Can you please not run the code through a formatter? If the code needs to be formatted it deserves a PR of its own, where the settings of the formatter can be discussed and more importantly - where there is no need to go through all the code changes because it's auto-generated anyways and contains no real changes. When you do it as part of a regular PR, most of the changes are formatting changes which makes it very hard to review the real changes.
> 
> Sure, and sorry about that. Not using the auto-formatter at all makes it pretty difficult for me personally to work on code, but I'll add a commit that reverts those changes once I take care of the other problems with my changes.

I don't mind setting up a formatter configuration and formatting all the code - as long as it's a separate PR. Or I can do it myself, if you don't mind handling the merge conflicts in this PR. Personally I prefer to set up a linter and do what it says because I like my line breaks at logical places and formatters tend to blindly re-format macros and function arguments according to width, but that can probably turned off in the configuration.
 
> > * I'm not sure `strip_collection` is the right term. I think `extend` better conveys what t his feature does (each call to the setter extends the collection)
> 
> It's a bit more complicated, since (as far as I can tell) collections aren't thoroughly defined by a single trait.
> Right now I'm using `Extend` for repeat calls, but `IntoIterator::Item` to strictly constrain the argument on the first call and `FromIterator` for the default collection initialiser if no other default is present. This mirrors how I'd hand-write such a builder.
> 
> It's definitely possible to reduce this to just `Extend`, but I'm concerned that would hurt ergonomics and could have a slight performance impact.

Maybe we are looking at it wrong. What if instead of building the end collection, the builder will always build a `Vec` and convert it on `build()` to whatever 

> > * Is it even required? Does it have real use-cases? I understand it can be used to start a collection with already existing entries, but every other field type requires the user to either use the default or set the entire content of the field. We don't have a setting for string fields to prefix the set strings with something...
> 
> It's an edge case, I think. For the default collections in `std` it's not necessary at all.
> 
> Collections that by type require at least one entry are more ergonomic in some use-cases, in that you don't have to `.unwrap()` any return values from e.g. `.first()` and `.last()`. This also means they mostly can't implement `Default` and usually won't implement `FromIterator`, though.
> There doesn't seem to be a standard way to move a single item into a new collection, which makes configuration necessary.

I think I kind of missed that `v1_first` is the item passed to the first call to `v1()`. I thought it was some global const or something, and that your `v1` will always have two members - `v1_first` and the argument of the first call. Which seems weird, unless you want to demonstrate an example and then it's just weird to have that feature in the first place.

That's what I get for skimming a PR, I guess, because if the initializer creates it the collection from the first item then it does make sense.

May I suggest to use a closure instead? So instead of

```rust
#[builder(setter(extend(initializer = vec![v1_first])))]
v1: Vec<i32>, 
```

You'd have

```rust
#[builder(setter(extend(initializer = |first| vec![first])))]
v1: Vec<i32>, 
```
This makes it more clear that an argument is involved.

> ```rust
> #[builder(default, setter(strip_option, extend(initializer = vec1![entry_first])))]
> entry: Option<Vec1<Entry>>,
> ```

image

Tamschi:
(I'm going to collapse a few of the nested quotes for brevity's sake.)

> > […]
> 
> I don't mind setting up a formatter configuration and formatting all the code - as long as it's a separate PR. Or I can do it myself, if you don't mind handling the merge conflicts in this PR. Personally I prefer to set up a linter and do what it says because I like my line breaks at logical places and formatters tend to blindly re-format macros and function arguments according to width, but that can probably turned off in the configuration.

I'd be glad if you did, thank you. I was actually considering submitting a `rustfmt.toml` that changes the existing code as little as possible, but it's better if you choose the settings.

Unfortunately, I don't think it's possible to make rustfmt preserve argument formatting. (You can set a maximum width, but that's it.)
I did the `cargo fmt` in a separate commit at the very start to make it easy to revert, so even if there won't be a formatter configuration, it won't cause extra work for me here.

> > > * I'm not sure `strip_collection` is the right term. I think `extend` better conveys what t his feature does (each call to the setter extends the collection)
> > 
> > 
> > It's a bit more complicated, since (as far as I can tell) collections aren't thoroughly defined by a single trait.
> > Right now I'm using `Extend` for repeat calls, but `IntoIterator::Item` to strictly constrain the argument on the first call and `FromIterator` for the default collection initialiser if no other default is present. This mirrors how I'd hand-write such a builder.
> > It's definitely possible to reduce this to just `Extend`, but I'm concerned that would hurt ergonomics and could have a slight performance impact.
> 
> Maybe we are looking at it wrong. What if instead of building the end collection, the builder will always build a `Vec` and convert it on `build()` to whatever

It's definitely possible, yes. This would then use only `FromIterator` and `IntoIterator::Item` by default, or maybe no specific trait at all with a custom initializer.

I see three drawbacks with this:

- Slightly lower performance, since the values would (likely) be copied through the heap first in all cases.

  I consider this only a minor issue since the library doesn't advertise runtime speed. It's hard to tell whether this would be noticeable with my framework, too, since this probably wouldn't be super frequent even if it does appear in the render path.

- A limitation on one item type, since `Extends` could be overloaded quite a bit. [Syn's `Punctuated` is an example of this](https://docs.rs/syn/1.0.58/syn/punctuated/struct.Punctuated.html), but maybe that (generally speaking) is out of scope?

- Delayed panics until `.build()` is called. Some collection types may accept only a certain range of values or panic on duplicate keys. Only panicking on `.build()` in that case seems like it would be harder to debug.

  I consider this the most serious problem, but it's still something that wouldn't come up very often, I think.

> > > * Is it even required? Does it have real use-cases? I understand it can be used to start a collection with already existing entries, but every other field type requires the user to either use the default or set the entire content of the field. We don't have a setting for string fields to prefix the set strings with something...
> > 
> > 
> > It's an edge case, I think. For the default collections in `std` it's not necessary at all.
> > Collections that by type require at least one entry are more ergonomic in some use-cases, in that you don't have to `.unwrap()` any return values from e.g. `.first()` and `.last()`. This also means they mostly can't implement `Default` and usually won't implement `FromIterator`, though.
> > There doesn't seem to be a standard way to move a single item into a new collection, which makes configuration necessary.
> 
> I think I kind of missed that `v1_first` is the item passed to the first call to `v1()`. I thought it was some global const or something, and that your `v1` will always have two members - `v1_first` and the argument of the first call. Which seems weird, unless you want to demonstrate an example and then it's just weird to have that feature in the first place.
> 
> That's what I get for skimming a PR, I guess, because if the initializer creates it the collection from the first item then it does make sense.
> 
> May I suggest to use a closure instead? So instead of
> 
> ```rust
> #[builder(setter(extend(initializer = vec![v1_first])))]
> v1: Vec<i32>, 
> ```
> 
> You'd have
> 
> ```rust
> #[builder(setter(extend(initializer = |first| vec![first])))]
> v1: Vec<i32>, 
> ```
> 
> This makes it more clear that an argument is involved.
> 
> > ```rust
> > #[builder(default, setter(strip_option, extend(initializer = vec1![entry_first])))]
> > entry: Option<Vec1<Entry>>,
> > ```
> 
> image

image
Noted!

In hindsight that was really confusing syntax, yeah.
<details><summary>I'll make sure to move away from this in my other code too.</summary>

Event handlers in my web framework currently get a `self` implicitly, but I may move away from that when I rewrite the parser to ignore syntax errors there. The event argument itself will definitely be explicit once I make it available in the rework, and I'll have to look into hiding the DI node and bump allocator handle from user code if they aren't already, or at least namespace them somehow…

</details>

Thank you! This is very helpful.
I'm trying to take a programming break for a few days, but you should see updates again around the weekend.

idanarye:
I've added a `rustfmt.toml` and formatted the code. Could you please rebase to remove all the formatting changes noise?

Tamschi (commented):
Thank you and done 👍
I'll most likely take a break for a few more days, but for now I've added a few comments where I think the changes could use some explanation.

There might be a bit much copy-paste right now. I tend to clean that up fairly late into my features, but I'll have to see how it turns out in this case.

src/field_info.rs
```diff
@@ -291,7 +371,7 @@ impl SetterSettings {
                     Err(Error::new_spanned(expr, "Expected simple identifier".to_owned()))
                 }
             }
-            _ => Err(Error::new_spanned(expr, "Expected (<...>=<...>)")),
+            _ => Err(Error::new_spanned(expr, "Expected (<...>=<...>) or (<...>(<...>))")),
```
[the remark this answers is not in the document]
Tamschi: Maybe using [Lookahead1](https://docs.rs/syn/1.0.59/syn/parse/struct.Lookahead1.html) would make sense here, since that can generate a similar error message automatically. The syntax isn't as clean as the match statement here, though.

Emilgardis:
Just a fyi, rustfmt will skip items marked `#[rustfmt::skip]`

Tamschi:
I haven't been able to work on this much lately, but here's a little progress.

Accepting `IntoIterator`s with the same name as the field and single items with a separate name should be much more intuitive.
I went with -`_item` as default suffix, but this is configurable through the `item_name` parameter.

Just `extend` will implement everything with default settings, while `extend()` (with empty parentheses) will generate only the methods that extend an existing collection, which makes it impossible to set the field. Adding `from_first` or `from_iter` as custom keywords into the parentheses will add the default implementation (which extends a specified default if present), assigning lambdas to them will use a custom implementation as before.

I'll have to think a bit about how to integrate this with `strip_option`. The cleanest version might be to strip the option also from the builder field type (only in combination with `extend`? Changing it in general could break things downstream.), wrapping it in `Some(…)` only in `.build()`.

tests/extend.rs
```diff
@@ -0,0 +1,67 @@
...
+fn extend_default_explicit_auto() {
+    #[derive(TypedBuilder)]
+    struct A {
+        #[builder(default = vec![0], setter(extend(from_first, from_iter)))]
+        v: Vec<i8>,
+    }
+
+    assert_eq!(A::builder().v_item(2).build().v, vec![0, 2]);
+    assert_eq!(A::builder().v(vec![3, 4]).build().v, vec![0, 3, 4]);
+}
```
Tamschi, line 55: Writing this out in the tests made me realise this is probably unexpected behaviour (and would also combine awkwardly with `strip_option`). I'll change it so that the auto-implementations always start with only the given items.

src/struct_info.rs
```diff
@@ -265,64 +265,247 @@ impl<'a> StructInfo<'a> {
...
-                        _phantom: self._phantom,
+                let input = match inputs.len() {
+                    0 => return Err(Error::new_spanned(quote!(#or1_token #or2_token), "Expected one argument")),
+                    1 => inputs.into_iter().next().unwrap(),
+                    _ => return Err(Error::new_spanned(inputs, "Expected only one argument")),
+                };
+
+                if let syn::Pat::Type(syn::PatType { colon_token, ty, .. }) = input {
+                    return Err(Error::new_spanned(quote!(#colon_token #ty), "Did not expect argument type, since it is set explicitly by the macro"))
+                }
```
Tamschi, line 346: I'll likely revisit this, making a closure argument type override the outer method argument type verbatim (including any `impl`).
That isn't quite standard Rust (though Syn should still parse it), but is needed to constrain the argument type for e.g. non-empty target collections.

src/field_info.rs
```diff
@@ -40,41 +40,61 @@ impl<'a> FieldInfo<'a> {
...
-            return None;
-        };
-        if let syn::GenericArgument::Type(ty) = generic_params.args.first()? {
-            Some(ty)
-        } else {
-            None
         }
+
+        try_(self).ok_or_else(|| Error::new_spanned(&self.ty, "can't `strip_option` - field is not `Option<...>`"))
     }
```
Tamschi, line 98: I ended up using this method in a few more places, so I unified the error handling here.
This also means I changed a few other functions slightly to propagate these errors, but that should be fairly minimal.

src/struct_info.rs
```diff
@@ -609,7 +605,10 @@ impl<'a> StructInfo<'a> {
             ),
             proc_macro2::Span::call_site(),
         );
-        let early_build_error_message = format!("Missing required field {}", field_name);
+        let mut early_build_error_message = format!("Missing required field {}", field_name);
+        if field.builder_attr.setter.extend.is_some() {
+            write!(&mut early_build_error_message, " (single item setter: {})", field.item_name()).unwrap();
+        }
```
Tamschi, line 611: There probably should be a special message in case neither of these are available as initialisers.
I'll stop for today, but I've added this point to my to-do list.

(As an aside, I've just been able to confirm this feature will work for my use-case 🥳
I've queued that feature up just so it's easier to keep track of, but that won't change how I work on this one.)

examples/example.rs
```diff
@@ -13,16 +23,73 @@ struct Foo {
...
     z: i32,
+
+    // #[builder(default)] without parameter - don't require this field
+    // #[builder(setter(strip_collection))] without parameter - start with the default and extend from there
+    #[builder(default, setter(strip_collection))]
+    v0: Vec<i32>,
+
+    // No `default`: This field must be set at least once.
+    // You can explicitly create the collection from the first item (but this is not required even without `default`).
+    #[builder(setter(strip_collection(from_first = |first| vec![first])))]
```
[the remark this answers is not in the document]
idanarye: Is `from_first` without arguments even meaningful? What would be the default?

As for `from_iter` - I think maybe it should be **the** default - as in, without an argument you won't even need to specify it. Instead of keeping the actual target collection we can keep `I: Iterator`, and each time we extend with the builder, we can make an [`std::iter::Chain`](https://doc.rust-lang.org/std/iter/struct.Chain.html). We can still specify `from_iter = |it| ...` if `FromIterator` is not available for that collection or if for whatever reason we need custom behavior.

Tamschi:
A small status update since it's been a month: A few things with higher priority have been happening, so I haven't found time to continue here. I currently don't have an ETA on when I'll be able to resume work on this draft.

Tamschi:
I should probably put a small status update here so that it doesn't look like I ran off as soon as this started working for my specific use case…

The reality of the situation is that I've been swamped with higher-priority work pretty much constantly since April, so I haven't found time to continue work on this feature here or the grammar feature in my framework that I'll need this for.
I'll likely be able to maintain this feature branch to make sure it doesn't fall behind over time, but I don't have an estimate on when I'll get around to completing it. Sorry about that.

idanarye:
No need to apologize. This is open source - we all work for free here, and no one should feel pressured to prioritize contribution over other things they have in their life.

Also - it's not like I've been working on this crate that much this year...

PR #54, opened 2021-07-17
Implemented setter(transform), resolves #53

### LINKED ISSUE #53: Feature request: Generalize `strip_option`
Generalize `strip_option` to a generic attribute accepting a method and an argument list, which the setter method would proxy to.

## Example
To build a struct with a `Cell`, we do not want to require the user to construct the `Cell` themselves.

```rust
#[derive(TypedBuilder)]
struct Foo {
    #[builder(transform = Cell::new(bar: u32))]
    bar: Cell<u32>,
}
```

The new attribute here is (in decl macro notation) `transform = $path:path( $($name:pat : $ty:ty),* )`.
A builder method is generated to require the specified parameters `$name*` and pass them to the function in `$path`, whose return value is used to set the field.

An equivalent notation for `strip_option` is `transform = ::std::option::Option::Some(field: Type)`.

## Alternatives
Use a trait to standardize this to avoid attribute parsing Example:

```
trait BuilderParam<U> : Sized {
    fn convert(self) -> U;
}
```

Then applying `#[builder(transform)]` on `field: Type` would generate

```rust
fn field(mut self, field: impl BuilderDelegate<Type>) -> Builder<...> {
    Builder {
        field: field.convert(),
        ...self
    }
}
```

I can write a patch for this if you find this appropriate.

SOF3:
Using the `#[builder(setter(transform = |a: T, b: U| f(a, b)))]` syntax, resolves #53

### DIFF
--- a/src/field_info.rs
+++ b/src/field_info.rs
@@ -90,6 +90,7 @@ pub struct SetterSettings {
     pub skip: bool,
     pub auto_into: bool,
     pub strip_option: bool,
+    pub transform: Option<Transform>,
 }
 
 impl FieldBuilderAttr {
@@ -231,6 +232,11 @@ impl SetterSettings {
                         self.doc = Some(*assign.right);
                         Ok(())
                     }
+                    "transform" => {
+                        self.transform = Some(parse_transform_closure(&assign.right)?);
+                        self.strip_option = false;
+                        Ok(())
+                    }
                     _ => Err(Error::new_spanned(&assign, format!("Unknown parameter {:?}", name))),
                 }
             }
@@ -296,4 +302,42 @@ impl SetterSettings {
             _ => Err(Error::new_spanned(expr, "Expected (<...>=<...>)")),
         }
     }
+}
+
+#[derive(Debug, Clone)]
+pub struct Transform {
+    pub params: Vec<(syn::Pat, syn::Type)>,
+    pub body: syn::Expr,
+}
+
+fn parse_transform_closure(expr: &syn::Expr) -> Result<Transform, Error> {
+    let closure = match expr {
+        syn::Expr::Closure(closure) => closure,
+        _ => return Err(Error::new_spanned(expr, "Expected closure")),
+    };
+    if let Some(kw) = &closure.asyncness {
+        return Err(Error::new(kw.span, "Transform closure cannot be async"));
+    }
+    if let Some(kw) = &closure.capture {
+        return Err(Error::new(kw.span, "Transform closure cannot capture variables"));
+    }
+
+    let inputs = &closure.inputs;
+    let mut params = Vec::new();
+    for input in inputs {
+        let param = match input {
+            syn::Pat::Type(pat_type) => {
+                (syn::Pat::clone(&pat_type.pat), syn::Type::clone(&pat_type.ty))
+            }
+            _ => return Err(Error::new_spanned(input, "Transform closure must explicitly declare types")),
+        };
+        params.push(param);
+    }
+
+    let body = &closure.body;
+
+    Ok(Transform {
+        params,
+        body: syn::Expr::clone(body),
+    })
 }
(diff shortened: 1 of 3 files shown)


src/field_info.rs
```diff
@@ -297,3 +303,41 @@ impl SetterSettings {
...
+fn parse_transform_closure(expr: &syn::Expr) -> Result<Transform, Error> {
+    let closure = match expr {
+        syn::Expr::Closure(closure) => closure,
+        _ => return Err(Error::new_spanned(expr, "Expected closure")),
+    };
+    if let Some(kw) = &closure.asyncness {
+        return Err(Error::new(kw.span, "Transform closure cannot be async"));
+    }
+    if let Some(kw) = &closure.capture {
+        return Err(Error::new(kw.span, "Transform closure cannot capture variables"));
```
idanarye, line 322: Nit: This error message is a bit confusing at first, because syn calls it `capture` but from the user's view the more natural term is `move`. It's not that bad, because after thinking about it for 10 seconds you can understand what it means, because `rustc` will mark the `move`, and because there is no reason to put `move` on that closure in the first place. Still - I will change that error message if this PR gets merged before you do it.

src/field_info.rs
```diff
@@ -297,3 +303,41 @@ impl SetterSettings {
...
+    }
+
+    let inputs = &closure.inputs;
+    let mut params = Vec::new();
+    for input in inputs {
```
idanarye, line 327: This can be done with a `collect`.

src/lib.rs
```diff
@@ -101,7 +102,14 @@ mod util;
 ///        // Defaults to Some(13), option-stripping is performed:
 ///        #[builder(default = Some(13))]
 ///        z: Option<i32>,
+///
+///        // Accepts params `(x: f32, y: f32)`
+///        #[builder(setter(transform = |x: f32, y: f32| Point { x, y }))]
+///        w: Point,
 ///    }
+///
+///    #[derive(Default)]
```
idanarye, line 111: I get this this is just an example, but... why?

src/lib.rs
```diff
@@ -134,6 +142,10 @@ mod util;
 ///     `Some(...)`, relieving the caller from having to do this. Note that with this setting on
 ///     one cannot set the field to `None` with the setter - so the only way to get it to be `None`
 ///     is by using `#[builder(default)]` and not calling the field's setter.
+///
+///   - `transform = |params: Types, ...| expr`: this makes the setter accept `params: Types, ...`
```
idanarye, line 146: This is a weird way to document that `transform` supports multiple parameters. How about `|param1: Type1, param2: Type2, ...|`?

src/lib.rs
```diff
@@ -134,6 +142,10 @@ mod util;
...
+///     instead of the field type itself. The parameters are transformed into the field type using
+///     the expression `expr`.
```
idanarye, line 148: Since the transformation may have side-effects, I think it may be worthwhile to mention here that it is done when calling the setter (as opposed to when calling `build`)

src/field_info.rs
```diff
@@ -231,6 +232,11 @@ impl SetterSettings {
                         self.doc = Some(*assign.right);
                         Ok(())
                     }
+                    "transform" => {
```
idanarye, line 235: `transform` cancels `strip_option`, so I think we should make it illegal to use put both on the same field. This can be tricky though (requires a change to the `handle_fields` macro) so I might do it myself after merging.

src/field_info.rs
SOF3: Since the user cannot capture any variables anyway, my motivation is that any such keywords would not make sense either way.

src/field_info.rs
SOF3: I used a loop because I didn't want to deal with `collect::<Result>()`. Could change it if it's a problem for you.

src/lib.rs
SOF3: Because all fields are `default` by default, and I intend to change as little code as possible.

src/field_info.rs
SOF3: The automatic cancellation is intended for the `field`defaults` one. Explicitly specifying both in the field attribute would be unreasonable, but I found it not so easy to distinguish, so I left it alone for the while.

PR #72, opened 2022-10-28
feat: add struct attr to replace `build` method with a private method 

fMeow:
Instead of generating a public `build` method, we can generate a private `build` method with a name like `__build()`. With this, users can write a custom build method with complicated logic and even build into another `struct`.  
### Normal case: generate pub `build` method

Say we have a struct `Foo` as following:
```rust
#[derive(PartialEq, TypedBuilder)]
struct Foo {
    x: i32,

    #[builder(default, setter(strip_option))]
    y: Option<i32>,

    #[builder(default = 20)]
    z: i32,
}
```
This is taken from `example/example.rs`.

The above code will generated a build method as following:

```rust
impl<
    __z: FooBuilder_Optional<i32>,
    __y: FooBuilder_Optional<Option<i32>>,
> FooBuilder<((i32,), __y, __z)> {
    pub fn build(self) -> Foo {
        let (x, y, z) = self.fields;
        let x = x.0;
        let y = FooBuilder_Optional::into_value(
            y,
            || ::core::default::Default::default(),
        );
        let z = FooBuilder_Optional::into_value(z, || 20);
        Foo { x, y, z }
    }
}
```
However, this default `build` implementation is quite simple and impossible to build into another `struct`.

### Generate a private `__build` method and impl custom build method
Now we set the new struct attr `build_method`:

```rust
#[derive(PartialEq, TypedBuilder)]
#[builder(build_method(vis="pub(crate)", name=__build))) // HERE
struct Foo {
    x: i32,

    #[builder(default, setter(strip_option))]
    y: Option<i32>,

    #[builder(default = 20)]
    z: i32,
}
```
The above code will generated a private method as following:
```rust
impl<
    __z: FooBuilder_Optional<i32>,
    __y: FooBuilder_Optional<Option<i32>>,
> FooBuilder<((i32,), __y, __z)> {
    pub(crate) fn __build(self) -> Foo { // Here is the difference
        let (x, y, z) = self.fields;
        let x = x.0;
        let y = FooBuilder_Optional::into_value(
            y,
            || ::core::default::Default::default(),
        );
        let z = FooBuilder_Optional::into_value(z, || 20);
        Foo { x, y, z }
    }
}
```

Now we can impl custom build method:

```rust
#[allow(non_camel_case_types)]
impl<
    __z: FooBuilder_Optional<i32>,
    __y: FooBuilder_Optional<Option<i32>>,
> FooBuilder<((i32,), __y, __z)> {
    pub fn build(self) -> Result<Bar, std::io::Error> {
        let foo = self.__build();
        // add custom code, whether to check or transform data
        Ok(Bar { foo.x, foo.y, foo.z })
    }
}
```
Note that the above code is not generated by this macro and we have to manually write these boilerplate code. We don't need to infer these generics ourself, we can use `cargo expand` to copy the implementation of `__build()` and modify it.

## On custom build method

Without this PR, we can also use this pattern to generate custom method to build struct. But the `build` method is already defined and we have to use another name like `finish()` or so. In this way, we have both `build()` and `finish()`, which can cause confusion. Even more so if the custom `finish` share a same signature with `build()`, in which we cannot detect the problem in compile time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `#[builder(build_method(vis="pub", name=build))]` for customizing visibility and fn name of the final build method
+  (the default visibility is `pub`, and default build name is `build`)
 
 ## 0.10.0 - 2022-02-13
 ### Added
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Foo::builder().x(1).y(2).y(3); // y is specified twice
 * Compile time verification that no field is set more than once.
 * Ability to annotate fields with `#[builder(default)]` to make them optional and specify a default value when the user does not set them.
 * Generates simple documentation for the `.builder()` method.
+* Customizable method name and visibility of the `.build()` method.
 
 ## Limitations
 
--- a/examples/complicate_build.rs
+++ b/examples/complicate_build.rs
@@ -1,0 +1,54 @@
+mod scope {
+    use typed_builder::TypedBuilder;
+
+    #[derive(PartialEq, TypedBuilder)]
+    #[builder(build_method(vis="", name=__build))]
+    pub struct Foo {
+        // Mandatory Field:
+        x: i32,
+
+        // #[builder(default)] without parameter - use the type's default
+        // #[builder(setter(strip_option))] - wrap the setter argument with `Some(...)`
+        #[builder(default, setter(strip_option))]
+        y: Option<i32>,
+
+        // Or you can set the default
+        #[builder(default = 20)]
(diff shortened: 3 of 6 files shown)


idanarye:
The general style in this crate's attributes is to use nesting instead of long names. So I'd prefer if the syntax would be like this:

```rust
#[derive(PartialEq, TypedBuilder)]
#[builder(private_build_method))
#[builder(build_method(vis = "pub(crate)", name = __build))
struct Foo {
    // ...
```
This will also allow setting the method's visibility to whatever you want (private should be `vis = ""`) and the name to whatever you want.

Also, please document this in `derive_typed_builder`'s doc comment. The doc comments on `TypeBuilderAttr` are not enough because the crate does not expose that struct (not that it could, seeing how it's a proc macro crate)

And update the changelog, too.

fMeow:
I have update this PR to add `#[builder(build_method(vis = "...", name = ...))` instead.

Also, I add an example to show how to use this pattern to create a more versatile build method.

Issue #77, opened 2023-03-03
Set default type for generated Builder generic


FlixCoder:
Hi!
Thank you for the great work on the TypedBuilder, it is really nice to use.

I have quite big structs and therefore made a wrapper that boxes the contents, so that it can be passed around without causing stack overflows. I.e.:

```rust
#[derive(TypedBuilder)]
struct Inner {
    a: i32,
    b: i32,
}

struct Outer(Box<Inner>);
```

The issue now is, that I want to build the outer type like this:
```rust
let value = Outer::builder().a(3).b(5).build();
```

There is no way for a custom build function and that can be worked around by adding a function `Inner::outer(self) -> Outer`, so that would be acceptable, as it would result in:
```rust
let value = Outer::builder().a(3).b(5).build().outer();
```

However, it is not feasible to create the `Outer::builder()` function, as the generic type needs to be set to a tuple of empty tuples, given the number of fields. The number of fields is sadly unknown, so I cannot generate the builder function appropriately.

So I was wondering if it was possible to somehow provide the default type of the builder via a default type parameter (not sure if that works) or a trait (that should be easy, but a bit clunky). `TypedBuilder` would generate this:

1. default type parameter:
```rust
struct BlaBuilder<TypedBuilderFields = ((), (), ...)> { ... }
```
2. trait:
```rust
pub trait BuilderDefaultType {
    type DefaultType;
}

impl<R> BuilderDefaultType for BlaBuilder<R> {
    type DefaultType = ((), (), ...);
}
```

Then I could use it like this hopefully?:
```rust
impl Outer {
    // 1
    pub fn builder() -> InnerBuilder {
        Inner::builder()
    }

    // 2
    pub fn builder() -> <InnerBuilder as BuilderDefaultType>::DefaultType {
        Inner::builder()
    }
}
```

What do you think about this, is this something that is possible and would be accepted? Especially variant 1 should not really change anything besides adding the default type parameter and should be easy to add without any harm, right?

FlixCoder:
I guess the `.outer()` function can be replaced by the solution of https://github.com/idanarye/rust-typed-builder/issues/52 ?

Issue #52, opened 2021-04-15
feature request: Custom `build` method


lebensterben:
## Status quo

Currently, if we have a struct `Foo` that derive `TypedBuilder`, it will generate something like
```rust
impl Foo {
  pub fn builder() -> FooBuilder<...>;
}
impl FooBuilder {
  pub fn build(self) -> Foo;
}
```

## Inconvenience

But there are some use cases that `Foo` is just an intermediate type, or `Foo` is itself a builder.

With `derive-builder`, it's possible to have, for example
```rust
pub struct Foo { /* fields omitted */ };
pub struct FooBuilder {
  foo: u8,
  bar: u8,
}
```
where `FooBuilder` derive the builder pattern. But user can override the `build()` function, so that
something like
```
FooBuilder::default().foo(1).bar(2).build()?
```
Generates a `Result<Foo, ..>` instead of `Result<FooBuilder, ..>`.

I understand that the mechanism of typed builder is different. And to achieve the same effect, I need to write something like
```rust
impl FooBuilder {
  pub fn build_foo(self) -> Result<Foo, ..>;
}
```
and
```
FooBuilder::builder().foo(1).bar(2).build().build_foo()?
```

## Proposal

I suggest to add a struct level attribute, `builder_into`, for example
```rust
pub struct Foo { /* fields omitted */ };

#[derive(TypedBuilder)]
#[builder(build_into = "Foo")]
pub struct FooBuilder { /* fields omitted */ };

impl From<FooBuilder> for Foo {
  fn from(foo_builder: FooBuilder) -> Foo;
}
```

And if `build_into` is empty, `FooBuilderBuilder::build()` should maintain the current behaviour.
And if it's not, as in the case above, it should behave like
```rust
<FooBuilder as Into<Foo>>::into(FoobuilderBuilder::build();
```

That is, the `FooBuilderBuilder::build()` method first construct the `FooBuilder` struct, and automatically convert it into `Foo` struct.

I think this is very helpful.
Thanks.

idanarye:
Counter offer:

```rust
pub struct Foo {
    // ...
}

#[typed_builder::function(execute=build)]
pub fn make_foo(
    fields: That,
    you: Would,
    have: Put,
    inside: FooBuilder,
) -> Foo {
    // Code for making Foo
}
```

`make_foo()` would create the builder.

If we want it as a static method, we would have to do something like that:

```rust
pub struct Foo {
    // ...
}

#[typed_builder::methods]
[...]

PR #79, opened 2023-03-03
feat: Set default builder type to empty builder

FlixCoder:
Fixes https://github.com/idanarye/rust-typed-builder/issues/77 .

It is a bit unclean yet, but I wanted to try if it works and it does.

The generic parameter needs to be at the end now, because generic parameters with a default cannot be before any paramter without a default.

### DIFF
--- a/src/struct_info.rs
+++ b/src/struct_info.rs
@@ -62,14 +62,16 @@ impl<'a> StructInfo<'a> {
             ..
         } = *self;
         let (impl_generics, ty_generics, where_clause) = self.generics.split_for_impl();
-        let all_fields_param =
-            syn::GenericParam::Type(syn::Ident::new("TypedBuilderFields", proc_macro2::Span::call_site()).into());
-        let b_generics = self.modify_generics(|g| {
-            g.params.insert(0, all_fields_param.clone());
-        });
         let empties_tuple = type_tuple(self.included_fields().map(|_| empty_type()));
+        let mut all_fields_param_type: syn::TypeParam =
+            syn::Ident::new("TypedBuilderFields", proc_macro2::Span::call_site()).into();
+        let all_fields_param = syn::GenericParam::Type(all_fields_param_type.clone());
+        all_fields_param_type.default = Some(syn::Type::Tuple(empties_tuple.clone()));
+        let b_generics = self.modify_generics(|g| {
+            g.params.push(syn::GenericParam::Type(all_fields_param_type.clone()));
+        });
         let generics_with_empty = modify_types_generics_hack(&ty_generics, |args| {
-            args.insert(0, syn::GenericArgument::Type(empties_tuple.clone().into()));
+            args.push(syn::GenericArgument::Type(empties_tuple.clone().into()));
         });
         let phantom_generics = self.generics.params.iter().map(|param| match param {
             syn::GenericParam::Lifetime(lifetime) => {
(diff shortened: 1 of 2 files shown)


src/struct_info.rs
```diff
@@ -62,14 +62,16 @@ impl<'a> StructInfo<'a> {
...
         let (impl_generics, ty_generics, where_clause) = self.generics.split_for_impl();
-        let all_fields_param =
-            syn::GenericParam::Type(syn::Ident::new("TypedBuilderFields", proc_macro2::Span::call_site()).into());
+        let empties_tuple = type_tuple(self.included_fields().map(|_| empty_type()));
+        let mut all_fields_param_type: syn::TypeParam =
+            syn::Ident::new("TypedBuilderFields", proc_macro2::Span::call_site()).into();
+        let all_fields_param = syn::GenericParam::Type(all_fields_param_type.clone());
         let b_generics = self.modify_generics(|g| {
-            g.params.insert(0, all_fields_param.clone());
+            all_fields_param_type.default = Some(syn::Type::Tuple(empties_tuple.clone()));
```
idanarye, line 70: Why does this need to be inside `modify_generics`?

src/struct_info.rs
```diff
@@ -62,14 +62,16 @@ impl<'a> StructInfo<'a> {
...
+            g.params.push(syn::GenericParam::Type(all_fields_param_type.clone()));
         });
-        let empties_tuple = type_tuple(self.included_fields().map(|_| empty_type()));
         let generics_with_empty = modify_types_generics_hack(&ty_generics, |args| {
-            args.insert(0, syn::GenericArgument::Type(empties_tuple.clone().into()));
+            args.push(syn::GenericArgument::Type(empties_tuple.clone().into()));
```
idanarye, line 74: Note to self: This is a breaking change and should be marked as such in the changelog.

tests/tests.rs
```diff
@@ -597,3 +597,31 @@ fn test_builder_type() {
...
+    #[derive(Debug, PartialEq)]
+    struct Outer(Inner);
+
+    impl Outer {
+        pub fn builder() -> InnerBuilder {
+            Inner::builder()
+        }
+    }
+
+    let outer = Outer::builder().a(3).b(5).build().outer();
```
idanarye, line 625: I really hate how `Outher::builder()` does not build an `outher`, but since it's not part of the actual API I'll allow it.

idanarye (approved):

src/struct_info.rs
idanarye: I just need an answer for that before I merge. I can move it outside after the merge, but I really don't want to leave it like that and forget about it...

FlixCoder: I was not sure what it might change if I do it outside, as it is reused in other places. Let me see if I can improve on this and see if it actually hurts. This is the part I meant when talking about it not being clean yet.

tests/tests.rs
FlixCoder: Yes, me too. I would really like it to return outer. But that needs some kind of #52, e.g. a possibility to set another type the builder outputs by using `.into()`.

I would really love something like this:
```rust
#[derive(TypedBuilder)]
#[builder(build_method(output = Outer))]
struct Inner { ... }
```
where `Outer` implements `From<Inner>` or something.. It is quite difficult to make your own build method or customizing it, so I don't really see a better way to allow such a pattern.

src/struct_info.rs
FlixCoder: Ah I see what you mean now. It does not need to be inside the closure, but it definitely needs to be after it was used to construct the `GenericParam::Type`. I am still not really happy with it, but outside is better.

idanarye: After I merge this PR, I think I'll just get rid of `modify_generics`. It was added by a very incompetent programmer (2018's me) and is not helpful enough to justify the confusion it sows.

PR #80, opened 2023-03-06
feat: Add into attribute to set the builder output

FlixCoder:
I implemented both my suggestions in one from #52, so you can take a look how it would look like. It allows both, the generic output that the user decides the build output and the explicit output type. By default nothing changes.

When setting `#[builder(into)]` without a value, the user needs to specify a type when building and the output is automatically turned into the specific value.
When setting `#[builder(into = Type)]`, the build method returns that type instead of the original. The compiler complains accordingly when the type does not implement `From<ActualType>`.

I personally like the results :) Let me know what you think.


src/struct_info.rs
```diff
@@ -574,11 +580,31 @@ impl CommonDeclarationSettings {
...
+    }
+}
+
 #[derive(Debug, Default)]
 pub struct TypeBuilderAttr {
     /// Whether to show docs for the `TypeBuilder` type (rather than hiding them).
     pub doc: bool,
 
+    /// Whether to convert the builded type into another while finishing the build.
+    pub into: IntoSetting,
```
idanarye, line 606: I think this belongs inside `build_method` settings, but given how they all currently use `CommonDeclarationSettings` it might be easier to keep it here for now, and I'll move it to `builder_method` after the merge (there is some work to be done with the attribute type, that I prefer to do myself)

tests/tests.rs
```diff
@@ -622,6 +616,24 @@ fn test_default_builder_type() {
...
+#[test]
+fn test_into_set_generic() {
+    #[derive(Debug, PartialEq, TypedBuilder)]
+    #[builder(into)]
+    struct SomeStruct<T: Default> {
+        #[builder(default)]
+        a: T,
+    }
+
+    let some: SomeStruct<bool> = SomeStruct::builder().build();
```
idanarye, line 638: Please add an `assert_eq` that verifies the value. This will also pacify the linter bot.

src/struct_info.rs
```diff
@@ -574,11 +580,31 @@ impl CommonDeclarationSettings {
...
 
+/// Setting of the `into` argument.
+#[derive(Debug)]
+pub enum IntoSetting {
+    /// Setting was not set.
+    Unset,
+    /// Setting was set without a value.
+    Set,
+    /// Setting was set with a specific value (type).
+    Type(syn::ExprPath),
```
idanarye, line 591: I don't like these variant names. They represent the syntactic way this setting is configured, and I think we should use names that represent the meaning of the setting. I suggest `NoConversion`, `GenericConversion` and `ConversionToSpecificType`.

idanarye (changes requested):

tests/tests.rs
FlixCoder: Oh whoops, forgot that, thanks

src/struct_info.rs
FlixCoder: Yeah I had my trouble with that, too. But I was also thinking, it might be symbolically still fine this way (as the whole builder will build into something different) and would allow to possibly modify the `builder()` method's visibility as well.

src/struct_info.rs
FlixCoder: Changed, thanks

FlixCoder:
By the way, do you know of the `darling` crate? It takes over the parsing of the attributed i to structs and could make plenty of code here obsolete I think. The downside is of course another dependency.

FlixCoder:
Oh, maybe a note in the changelog is missing as well or do you maintain that yourself?

FlixCoder:
Oh my god, I force pushed empty changes except the ones you suggested :( I am sorry

idanarye:
It's okay. They were very simple ones, and I was planning to do some modifications myself anyways.

tests/tests.rs
```diff
@@ -622,6 +616,24 @@ fn test_default_builder_type() {
...
+            Self(value)
+        }
+    }
+
+    let outer = Outer::builder().a(3).b(5).build();
     assert_eq!(outer, Outer(Inner { a: 3, b: 5 }));
 }
+
+#[test]
+fn test_into_set_generic() {
```
idanarye, line 630: @FlixCoder Just noticed this, but... how does this test the new feature exactly? `Into` into itself?

FlixCoder: It tests whether the generics are correctly set when generating the build function. The into works as per From trait and is basically negligble.

idanarye: But it basically converts from a type to itself - remove the `#[builder(build_method(into))]` and it'd still work.

I'm going to change it to use two type. Also, I'm going to make two versions - one that `impl From` and one that `impl Into`. I forgot that the standard Rust way is to check `Into` because there is blanket `impl` of `Into` if there is a `From`.

FlixCoder: The into specific type os in the test above with the outer already, but it is probably better to have a dedicated test right?

FlixCoder: Also the genericness requires From to be implemented, even though Into is used to convert, so it probably complain. I wonder if one can use `#name: Into<R>` instead to make it work.

idanarye: Yes, but it requires a `where` clause: https://github.com/idanarye/rust-typed-builder/commit/5cead25255bc7e9313cae40259b6e02172a3281a#diff-e8a945989e052d3a4f5c3366aa483e17330af013e0dacf70aed01f19da45237dR513

FlixCoder: Ah, nice catch, that is probably better

lebensterben:
I'm late to this discussion.

I think it'd be helpful if the new change is documented somewhere more discoverable.

idanarye:
@lebensterben How so? [It's documented in the docs](https://idanarye.github.io/rust-typed-builder/typed_builder/derive.TypedBuilder.html#customisation-with-attributes), like all other `#[builder(...)]`-activated features (except `default`, but without setting defaults there is little point in using the builder pattern in Rust so I do feel it deserves a place in the README). I'm not sure why `build_method(into)` needs to be more discoverable than the rest of them?

lebensterben:
@idanarye 

sorry. I overlooked changes to the lib.rs

idanarye:
No problem. Sometimes it does take time for them to update...

Issue #57, opened 2021-09-02
use of undeclared crate or module `proc_macro`


dblock:
Trying to use dry run in https://github.com/lycheeverse/lychee/pull/304 seeing this:

```
rustc --edition=2018 --crate-name typed_builder /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/typed-builder-0.9.0/src/lib.rs --color always --crate-type proc-macro --emit=dep-info,link -C prefer-dynamic -C debuginfo=2 -C metadata=72ebad225c590f6d -C extra-filename=-72ebad225c590f6d --out-dir /home/ubuntu/source/lychee/dblock-lychee/target/package/lychee-lib-0.7.1/target/debug/deps -L dependency=/home/ubuntu/source/lychee/dblock-lychee/target/package/lychee-lib-0.7.1/target/debug/deps --extern proc_macro2=/home/ubuntu/source/lychee/dblock-lychee/target/package/lychee-lib-0.7.1/target/debug/deps/libproc_macro2-6f5ad8ea0bdf5b2f.rlib --extern quote=/home/ubuntu/source/lychee/dblock-lychee/target/package/lychee-lib-0.7.1/target/debug/deps/libquote-9b9b7563b6863207.rlib --extern syn=/home/ubuntu/source/lychee/dblock-lychee/target/package/lychee-lib-0.7.1/target/debug/deps/libsyn-c3d8aa50fc12e127.rlib --cap-lints allow
```

```
   Verifying lychee-lib
error[E0433]: failed to resolve: use of undeclared crate or module `proc_macro`
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/typed-builder-0.9.0/src/lib.rs:138:36
    |
138 | pub fn derive_typed_builder(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
    |                                    ^^^^^^^^^^ use of undeclared crate or module `proc_macro`

error[E0433]: failed to resolve: use of undeclared crate or module `proc_macro`
   --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/typed-builder-0.9.0/src/lib.rs:138:64
    |
138 | pub fn derive_typed_builder(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
    |                                                                ^^^^^^^^^^ use of undeclared crate or module `proc_macro`

```

Ideas?

idanarye:
[...]

Issue #31, opened 2020-01-21
Split the proc macro to dependency crate


idanarye:
The `TypedBuidler` custom derive proc macro generates some helpers. In current master there are `{BuilderType}_Optional` and `{BuilderType}_core`, and #30 is adding some bottom types (`{BuilderType}_Error_Missing_required_field_{field_name}` and `{BuilderType}_Error_Repeated_field_{field_name}`)

These things are weirdly mangled and repeated for every struct with a typed builder even though they are exactly the same - it would be better if we could just put them in the typed builder crate, with legal Rust names, and just reuse them. The problem is that proc macro crates are not allowed to export anything that's not a proc macro.

The solution: split the proc macro to a new crate - say `typed-builder-procmacro` - and have `typed-builder` re-export it. This is the common solution for this problem in Rust-2018 where macros can be imported/exported like regular symbols.

idanarye:
Actually, I think I'll keep `{BuilderType}_Error_Repeated_field_{field_name}` - at least for now - because it gets into the build error. `{BuilderType}_Error_Missing_required_field_{field_name}` doesn't show in the error message so I'll replace it once I split the crate.

![image](https://user-images.githubusercontent.com/1149255/73107798-6e88c700-3f07-11ea-8124-3b0727e4e549.png)

idanarye:
Sure

PR #93, opened 2023-06-11
Extract macro to new crate

mo8it:
Updated the edition to 2021 and removed `extern crate proc_macro;` because it did not help at the end: https://github.com/idanarye/rust-typed-builder/issues/57#issuecomment-930505830


mo8it:
@idanarye I am not sure about what to do with the bottom types for repeated and required fields: https://github.com/idanarye/rust-typed-builder/issues/31#issuecomment-578324069

You said that you would like to remove the one for required fields and keep the one for repeated fields, but on the other hand, the error for repeated fields points to the method with the name of the field. Therefore, we could just replace both bottom types with `Missing_Required_Fields` and `Repeated_Field_Assignment`.

src/lib.rs
```diff
@@ -162,47 +148,23 @@ mod util;
...
-#[proc_macro_derive(TypedBuilder, attributes(builder))]
-pub fn derive_typed_builder(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
-    let input = parse_macro_input!(input as DeriveInput);
-    match impl_my_derive(&input) {
-        Ok(output) => output.into(),
-        Err(error) => error.to_compile_error().into(),
-    }
+pub use typed_builder_macro::TypedBuilder;
+
+#[doc(hidden)]
```
idanarye, line 153: Now that it's an official trait in its own crate, I think it should be exposed and documented. I can add the documentation myself later though.

typed-builder-macro/src/lib.rs
idanarye, line 1: This is mostly a direct copy-paste, right? (github won't show a proper diff because of the new `src/lib.rs`...)

idanarye:
This is weird, but now that I check again the repeated value error does not show the type in the rust-analyzer virtual text diagnostics and missing value does:

image

This is the opposite of what I've observed in #31. But it's actually better, because repeated value is easy to figure out even without the type name as the hint.

This means that I no longer mind having a `Repeated_Field_Assignment`, but I still want to generate `Missing_Required_Field_*`. Maybe when we get a const generic string we can use that instead.

src/lib.rs
mo8it: To be honest, I don't see a value in exposing the trait. Adding it to the documentation whould be a distraction since it is only used internally in the code generated by the macro.

But adding a normal comment would be nice.

mo8it:
@idanarye Thanks for merging!

We might need to adjust the CI if we add tests to the macro crate. I think we should use a workspace and add `--workspace` to the `cargo test` command. Should I submit a PR?

idanarye:
Why? All the tests are still main crate, not the proc-macro crate. They still run with a regular `cargo test`.

mo8it:
I just thought about the scenario of adding a test to subcrates that would not be run in the CI.

Aslong as we keep all tests there, there is no problem.

idanarye:
All the tests go in the `tests` directory at the project's root. I don't see why there should ever be a need to add some to the macro subcrate.

Issue #47, opened 2021-01-31
Macro panic with raw identifier


neilg:
I run into a panic in the derive macro with valid code using raw identifiers that are keywords. I'm generating code from an IDL document, and the generated code contains raw literals (r#identifier) to avoid conflicts with Rust keywords. The macro seems to work fine on raw literals except when they are keywords, e.g. r#async.

See the example below.

Expected: the code compiles.

Actual:
```
error: proc-macro derive panicked
 --> src/lib.rs:1:10
  |
1 | #[derive(typed_builder::TypedBuilder)]
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: `"__r#async"` is not a valid identifier
```

```rust
#[derive(typed_builder::TypedBuilder)]
pub struct Bob {
    r#this_is_fine: i32,
    r#async: i32,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn build_bob() {
        let bob = Bob::builder()
            .this_is_fine(1)
            .r#async(3)
            .build();
    }
}
```

Version (from Cargo.lock)
```
name = "typed-builder"
version = "0.8.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "2ec6661b8cd8dc76d17f15f1cf1da27fc3dd58850ccc220d04653b48e0fae3ba"
```

PR #94, opened 2023-06-16
feat: add support for setter prefix and suffix

Techassi:
This PR adds support to add prefixes and suffixes to setter methods.

```rust
#[derive(TypedBuilder)]
#[builder(field_defaults(setter(prefix = "with_", suffix = "_value")))]
struct Foo {
    x: i32,
    y: u32,
}

let foo = Foo::builder().with_x_value(1).with_y_value(2).build();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   setter for that field now does.
 
 ### Added
-- Support for setter method pre- and suffixes `#[builder(field_defaults(setter(prefix = "...", suffix = "...")))]`.
+- Support for setter method prefixes and suffixes `#[builder(field_defaults(setter(prefix = "...", suffix = "...")))]`.
   This either prepends or appends the provided string to the setter method. This allows method names like: `set_x()`,
   `with_y()`, or `set_z_value()`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   generated code itself does trigger the deprecation warning, and instead the
   setter for that field now does.
 
+### Added
+- Support for setter method pre- and suffixes `#[builder(field_defaults(setter(prefix = "...", suffix = "...")))]`.
+  This either prepends or appends the provided string to the setter method. This allows method names like: `set_x()`,
+  `with_y()`, or `set_z_value()`.
+
 ## 0.14.0 - 2023-03-08
 ### Added
 - `build_method(into)` and `build_method(into = ...)`.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,11 +150,11 @@
 ///     when the setter is called.
 ///
 ///   - `prefix = "..."` prepends the setter method with the specified prefix. For example, setting
-///     `prefix = "with"` results in setters like `with_x` or `with_y`. This option is combinable
+///     `prefix = "with_"` results in setters like `with_x` or `with_y`. This option is combinable
 ///     with `suffix = "..."`.
 ///
 ///   - `suffix = "..."` appends the setter method with the specified suffix. For example, setting
-///     `suffix = "value"` results in setters like `x_value` or `y_value`. This option is combinable
+///     `suffix = "_value"` results in setters like `x_value` or `y_value`. This option is combinable
 ///     with `prefix = "..."`.
 pub use typed_builder_macro::TypedBuilder;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,7 @@
 /// // Foo::builder().x(1).y(2).y(3);
 /// ```
 ///
-/// # Customisation with attributes
+/// # Customization with attributes
 ///
 /// In addition to putting `#[derive(TypedBuilder)]` on a type, you can specify a `#[builder(…)]`
 /// attribute on the type, and on any fields in it.
@@ -148,6 +148,14 @@
 ///     `param1: Type1, param2: Type2 ...` instead of the field type itself. The parameters are
 ///     transformed into the field type using the expression `expr`. The transformation is performed
 ///     when the setter is called.
+///
+///   - `prefix = "..."` prepends the setter method with the specified prefix. For example, setting
+///     `prefix = "with"` results in setters like `with_x` or `with_y`. This option is combinable
+///     with `suffix = "..."`.
+///
+///   - `suffix = "..."` appends the setter method with the specified suffix. For example, setting
+///     `suffix = "value"` results in setters like `x_value` or `y_value`. This option is combinable
+///     with `prefix = "..."`.
 pub use typed_builder_macro::TypedBuilder;
 
 #[doc(hidden)]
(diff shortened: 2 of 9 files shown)


mo8it:
May I ask why? It is just more to type repeatedly :/

Techassi:
It is nice to have the option. For example I personally like to prefix my (manually implemented) builder methods with `with_`.

This is no breaking change and users don't have to set these values. If they are not set, the setter method is named like the field (current behavior). So really users can choose if they want to type more.

mo8it:
I see your point, but I think that this subjective and adds complexity to the code. The whole crate and all its builder attributes have the purpose to type less.

I think that it is fine for a crate to be opionated to some degree.

But let's see what @idanarye thinks about it.

BTW: Always open an issue and wait for the response of the maintainers before starting with a PR. I was disappointed often enough by rejected PRs :)

idanarye:
While I myself don't see the point in using `set_` or `with_` prefixes in a builder that does not grant public access to the internal fields, Typed Builder already allows customizing a lot of things so I don't see a point to get opinionated about this thing in particular. And while it does add complexity being a new feature and all, I think that complexity is very localized.

typed-builder-macro/src/struct_info.rs
```diff
@@ -285,13 +287,76 @@ impl<'a> StructInfo<'a> {
...
                 )]
-                pub fn #field_name (self, _: #repeated_fields_error_type_name) -> #builder_name <#( #target_generics ),*> {
+                pub fn #method_name (self, _: #repeated_fields_error_type_name) -> #builder_name <#( #target_generics ),*> {
                     self
                 }
             }
         })
     }
 
+    fn method_name(field: &FieldInfo, field_name: &Ident) -> Result<Ident, Error> {
```
idanarye, line 297: Please change it to `setter_method_name`. While as a variable `method_name` makes sense inside the context of `field_impl`, it is pretty vague as a struct level method name. Especially when it's under `StructInfo`.

Speaking of which - why is this not under `FieldInfo`? And `FieldInfo` already has it's `name` inside it - there is no need to pass it twice.

typed-builder-macro/src/struct_info.rs
```diff
@@ -285,13 +287,76 @@ impl<'a> StructInfo<'a> {
...
+        let prefix = match &field.builder_attr.setter.prefix {
+            Some(prefix) => match prefix {
+                syn::Expr::Lit(lit) => match &lit.lit {
+                    syn::Lit::Str(str) => Some(syn::Ident::new(&format!("{}", str.value()), proc_macro2::Span::call_site())),
+                    _ => {
+                        return Err(Error::new_spanned(
```
idanarye, line 303: Wouldn't it make more sense to do this check in `SetterSettings::apply_meta`?

idanarye:
Please add a test with a keyword identifier for a field. Typed Builder had a problem with those in the past (#47), and I suspect this new feature (especially the prefix) may choke on it as well.

idanarye:
And please add an entry to the changelog.

typed-builder-macro/src/struct_info.rs
idanarye: Also, this whole thing (converting an `syn::Expr` to a string) should probably be extracted to a function.

typed-builder-macro/src/struct_info.rs
Techassi: Yes, that makes sense. Changed in cc1c5da

Techassi:
> Also, I'm not sure how I feel about the prefix and the postfix automatically adding the underscore.

Removed the underscore in cc1c5da



> And please add an entry to the changelog.

Updated in fbae931

Techassi:
> Please add a test with a keyword identifier for a field. Typed Builder had a problem with those in the past (#47), and I suspect this new feature (especially the prefix) may choke on it as well.

Added and fixed in ce48427

src/lib.rs
```diff
@@ -148,6 +148,14 @@
 ///     `param1: Type1, param2: Type2 ...` instead of the field type itself. The parameters are
 ///     transformed into the field type using the expression `expr`. The transformation is performed
 ///     when the setter is called.
+///
+///   - `prefix = "..."` prepends the setter method with the specified prefix. For example, setting
+///     `prefix = "with"` results in setters like `with_x` or `with_y`. This option is combinable
```
idanarye, line 153: The example here should say ``prefix = "with_"``. Similarly for the suffix.

typed-builder-macro/src/util.rs
```diff
@@ -85,6 +85,17 @@ pub fn strip_raw_ident_prefix(mut name: String) -> String {
     name
 }
 
+pub fn strip_raw_ident_prefix_with_status(mut name: String) -> (String, bool) {
+    let len = name.len();
+    name = strip_raw_ident_prefix(name);
+
+    if name.len() < len {
```
idanarye, line 92: 1. 
    ```rust
    (name, name.len() < len)
    ```
2. `strip_raw_ident_prefix` is simple enough that it'd be simpler and more straightforward to duplicate its code here instead of 
    * Alternatively - reverse the roles, so that `strip_raw_ident_prefix` calls `strip_raw_ident_prefix_with_status` and ditches the status.
4. A boolean output "named" `status` is not very descriptive.
5. This entire thing is actually redundant - see my comment in `setter_method_name`.

typed-builder-macro/src/field_info.rs
```diff
@@ -73,6 +76,26 @@ impl<'a> FieldInfo<'a> {
         }
     }
 
+    pub fn setter_method_name(&self) -> Ident {
+        let (name, was_stripped) = strip_raw_ident_prefix_with_status(self.name.to_string());
```
idanarye, line 80: You don't actually need `was_stripped`. You only need to re-add the `r#` if both the prefix and the suffix are `None`, but in that case you can just return `self.name` (`.clone()`)

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   generated code itself does trigger the deprecation warning, and instead the
   setter for that field now does.
 
+### Added
+- Support for setter method pre- and suffixes `#[builder(field_defaults(setter(prefix = "...", suffix = "...")))]`.
```
idanarye, line 14: Nit - "pre-" is confusing, and does not save much compared to just writing "prefixes".

typed-builder-macro/src/util.rs
Techassi: Removed in d7e98fc

typed-builder-macro/src/field_info.rs
```diff
@@ -73,6 +75,20 @@ impl<'a> FieldInfo<'a> {
...
+        let name = strip_raw_ident_prefix(self.name.to_string());
+
+        if let (Some(prefix), Some(suffix)) = (&self.builder_attr.setter.prefix, &self.builder_attr.setter.suffix) {
+            Ident::new(&format!("{}{}{}", prefix, name, suffix), Span::call_site())
+        } else if let Some(prefix) = &self.builder_attr.setter.prefix {
+            Ident::new(&format!("{}{}", prefix, name), Span::call_site())
+        } else if let Some(suffix) = &self.builder_attr.setter.suffix {
+            Ident::new(&format!("{}{}", name, suffix), Span::call_site())
+        } else {
+            Ident::new_raw(&name, Span::call_site())
```
idanarye, line 88: The point is that here you should return `self.name.clone()`, so that if the field name was `r#`'d the setter method name would be `r#`'d as well.

idanarye (changes requested):

typed-builder-macro/src/field_info.rs
Techassi: Oh man, sorry. Totally missed that. Will change asap.

Techassi: Changed in c38d686

PR #95, opened 2023-06-19
feat: Add PostBuild trait

### LINKED ISSUE #67: Allow invariant checks in `.build()`
I want something like this:

```rust
pub struct S {
    x: i32,
    y: i32,
}

impl S {
    check_rep(&self) {
        assert!(self.x < self.y);
    }
}
```

When `.build()` is called, I also want `check_rep` to be called.

Techassi:
Fixes #67

This PR adds support for the `PostBuild` trait, which allows users to perform custom validation when the `.build()` method is called. The validation runs as the last step of the `build` method.

An example:

```rust
use typed_builder::{PostBuild, TypedBuilder};

#[derive(Debug, PartialEq, TypedBuilder)]
#[builder(postbuild)]
struct Foo {
    x: i32,
    y: i32,
}

impl PostBuild for Foo {
    type Output = Result<Self, String>;

    fn postbuild(self) -> Self::Output {
        if self.x >= 5 {
            return Err("x too high - must be below or 5".into());
        }

        Ok(self)
    }
}

fn main() {
    let foo = Foo::builder().x(1).y(2).build().unwrap();
    assert_eq!(foo, Foo { x: 1, y: 2 });

    // Fails to validate during runtime
    // let foo = Foo::builder().x(5).y(6).build().unwrap();
    // assert_eq!(foo, Foo { x: 5, y: 6 });
}
``` 

When the user provides `#[builder(postbuild)]`, the macro **won't** implement a default `PostBuild` impl. Instead, the user has to implement this trait manually. This allows user to implement custom validation logic.

---

**Note:** Some test cases are commented out, as the initial draft of this feature breaks these test cases. Follow-up commits will fix those issues. As such, the PR is a draft. When the test cases are fixed and the code is polished, it is marked as "Ready to review".

### DIFF
--- a/examples/postbuild.rs
+++ b/examples/postbuild.rs
@@ -1,0 +1,29 @@
+use typed_builder::{PostBuild, TypedBuilder};
+
+#[derive(Debug, PartialEq, TypedBuilder)]
+#[builder(postbuild)]
+struct Foo {
+    x: i32,
+    y: i32,
+}
+
+impl PostBuild for Foo {
+    type Output = Result<Self, String>;
+
+    fn postbuild(self) -> Self::Output {
+        if self.x >= 5 {
+            return Err("x too high - must be below or 5".into());
+        }
+
+        Ok(self)
+    }
+}
+
+fn main() {
+    let foo = Foo::builder().x(1).y(2).build().unwrap();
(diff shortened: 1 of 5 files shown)


ZENOTME:
Hi, seems this pr has been a long time, I'm curious what is the status of this feature?

PR #97, opened 2023-06-22
Fix private type leak

AldaronLau:
This PR fixes a private type leak bug in the expanded macro output, by making a small simplification in the implementation.  I also added a test that triggers the bug.

### DIFF
--- a/tests/no_type_leakage.rs
+++ b/tests/no_type_leakage.rs
@@ -1,0 +1,27 @@
+// As long as this test compiles, it passes (test does not occur at runtime)
+
+use typed_builder::TypedBuilder;
+
+#[derive(PartialEq, TypedBuilder)]
+#[builder(
+    build_method(vis="pub", into=Bar),
+    builder_method(vis=""),
+    builder_type(vis="pub", name=BarBuilder),
+)]
+struct Foo {
+    x: i32,
+}
+
+pub struct Bar(Foo);
+
+impl Bar {
+    pub fn builder() -> BarBuilder {
+        Foo::builder()
+    }
+}
+
+impl From<Foo> for Bar {
+    fn from(wrapped: Foo) -> Self {
+        Bar(wrapped)
+    }
+}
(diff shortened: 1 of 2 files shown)


idanarye:
Can you explain what the bug is? I'm not sure I 100% follow.

AldaronLau:
The bug happens to when 

 1. The builder method is public
 2. The builder type is public and converts to a public type with `build_method(into =`
 3. The type being built is private
 4. At least one field is required / non-`default`

The generated code creates a method that produces a compiler error.

> ```
> error[E0446]: private type `Foo` in public interface
>   --> tests/no_type_leakage.rs:5:21
>    |
> 5  | #[derive(PartialEq, TypedBuilder)]
>    |                     ^^^^^^^^^^^^ can't leak private type
> ...
> 11 | struct Foo {
>    | ---------- `Foo` declared as private
>    |
>    = note: this error originates in the derive macro `TypedBuilder` (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> For more information about this error, try `rustc --explain E0446`.
> error: could not compile `typed-builder` (test "no_type_leakage") due to previous error
> ```

A snippet of expanded code showing where the error is:

```rust
#[doc(hidden)]
#[allow(dead_code, non_camel_case_types, non_snake_case)]
pub enum BarBuilder_Error_Missing_required_field_x {}
#[doc(hidden)]
#[allow(dead_code, non_camel_case_types, missing_docs, clippy::panic)]
impl BarBuilder<((),)> {
    #[deprecated(note = "Missing required field x")]
    pub fn build(self, _: BarBuilder_Error_Missing_required_field_x) -> Foo {
        ::core::panicking::panic("explicit panic");
    }
}
impl BarBuilder<((i32,),)> {
    pub fn build(self) -> Bar {
        let (x,) = self.fields;
        let x = x.0;
        #[allow(deprecated)] Foo { x }.into()
    }
}
```

As you can see, the missing required field `build()` method has the incorrect return type (`Foo`, when it should be `Bar`).  But, since it always panics, I opted to return `!` instead.

PR #111, opened 2023-08-14
Close #109: Add `crate_module_path`

### LINKED ISSUE #109: Generating #[derive(TypedBuilder)] in another crate forces all crates under to add typed-builder as a dependency
See https://github.com/leptos-rs/leptos/pull/1496. If Leptos updates to 0.15, it forces all leptos apps to add typed-builder to its dependencies in the Cargo.toml which is not ideal.

The easiest way to fix this is to revert this commit: https://github.com/idanarye/rust-typed-builder/commit/b9ca28737dd9309ea294df828583982f17a543d8, which would mean each type would now have its own Optional again. This would yield more generated code.

Had a note here, but I guess it was forgotten about:
https://github.com/idanarye/rust-typed-builder/commit/b9ca28737dd9309ea294df828583982f17a543d8#diff-f4519feb64908d4106578124d2bb7d0fbc21b38c3ae011d89aa95b2320feb7a8L170

Ideally, this should be resolved with https://github.com/rust-lang/rust/issues/54363. There are potential workarounds here: https://github.com/frozenlib/parse-display/issues/28, but that may require more effort to investigate and implement.

idanarye:


idanarye:
@awesomelemonade Before I merge, please check if this works for your usecase.

gbj:
I'm not @awesomelemonade but am maintainer of the Leptos framework that prompted the discussion, and yes, this should work perfectly, as it would allow us to specify the re-exported path to as `::leptos::typed_builder` rather than `::typed_builder`, so users don't have to add `typed_builder` to their own `Cargo.toml`. 

Thanks for all your work on this great library.

gbj:
It looks like there's an additional direct `::typed_builder` reference that might be missed in the PR, in the generics on the impl block.

For example, the following
```rust
#[derive(::leptos::typed_builder_macro::TypedBuilder)]
#[builder(crate_module_path = typed_builder)]
struct InfoGridProps<const NUM_ROWS: usize, const NUM_COLS: usize> {
    data: [[u32; NUM_ROWS]; NUM_COLS],
    #[builder(default)]
    optional_prop: bool,
}
```
expands to code including the following
```rust
impl<
    const NUM_ROWS: usize,
    const NUM_COLS: usize,
    // ❌ crate_module_path not used here
    __optional_prop: ::typed_builder::Optional<bool>,
> InfoGridPropsBuilder<
    NUM_ROWS,
    NUM_COLS,
    (([[u32; NUM_ROWS]; NUM_COLS],), __optional_prop),
> {
    pub fn build(self) -> InfoGridProps<NUM_ROWS, NUM_COLS> {
        let (data, optional_prop) = self.fields;
        let data = data.0;
        // crate_module_path is used here
        let optional_prop = ::leptos::typed_builder::Optional::into_value(
            optional_prop,
            || ::core::default::Default::default(),
        );
        #[allow(deprecated)]
        InfoGridProps {
            data,
            optional_prop,
        }
            .into()
    }
}
```

idanarye:
Yes, this is the problem:

https://github.com/idanarye/rust-typed-builder/blob/628786a5959ef7747cbb9f5423df80ebc9b0365f/typed-builder-macro/src/struct_info.rs#L404-L428

I think I missed it because, for performance reasons, it's not written as `::typed_builder::Optional`.

Not 100% sure if I should change it to a `syn::parse_quote!`...

idanarye:
@gbj I pushed a fix. Please try again.

gbj:
It works! Thank you so much!

PR #116, opened 2023-09-18
Implement mutators

### LINKED ISSUE #115: Consider supporting updating of already set values e.g. extending a Vec
My usecase is the following, I have a component with multiple fields, one of them is special, the children.

```rs
struct Component {
    normal_field: String,
    children: Html
}
```
I would like to use the builder to extend the Html, I created a workaround manually implementing a `child()` function:
```rs
impl<__normal_field>
    HtmlPageBuilder<(__normal_field, (Html,))>
{
    pub fn child(mut self, child: impl ToHtml) -> Self {
        child.write_to_html(&mut self.fields.1.0);
        self
    }
}

impl<__normal_field>
    HtmlPageBuilder<(__normal_field, ())>
{
    pub fn child(
        self,
        child: impl ToHtml,
    ) -> HtmlPageBuilder<(__mobile, __title, __styles, __scripts, (Html,))> {
        let (normal_field, ()) = self.fields;
        HtmlPageBuilder {
            fields: (normal_field, (child.into_html(),)),
            phantom: self.phantom,
        }
    }
}
```

I would prefer being able to have this built-in, maybe just as a parameter where I could pass an `Fn(&mut Html, impl IntoHtml)` or `Fn(Option<Html>, impl IntoHtml)`.

Something else I'd like would be to disable the default generated `children` function without removing the generic for `children`, i.e. something that was like `#[builder(setter(skip))]` but only for the setter and not for the generic.

ModProg:
closes #115.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,7 @@ use core::ops::FnOnce;
 ///    #[derive(Default)]
 ///    struct Point { x: f32, y: f32 }
 ///    ```
+///
 /// - `mutators(...)` takes functions, that can mutate fields inside of the builder.
 ///   See [mutators](#mutators) for details.
 ///
@@ -142,6 +143,9 @@ use core::ops::FnOnce;
 ///
 /// - `via_mutators = …` or `via_mutators(init = …)`: initialies the field with the expression `…`
 ///   when constructing the builder, useful in combination with [mutators](#mutators).
+///
+/// - `mutators(...)` takes functions, that can mutate fields inside of the builder.
+///   Mutators specified on a field, mark this field as required, see [mutators](#mutators) for details.
 ///
 /// - `setter(...)`: settings for the field setters. The following values are permitted inside:
 ///
@@ -190,6 +194,10 @@ use core::ops::FnOnce;
 /// fields can be specified via `self: (field1, field2)` in the mutator's function signature.
 /// Mutators that require non `via_mutators` fields, will only be availible if these fields are set.
 ///
+/// Mutators on a field, result in them automatically making the field required, i.e., its setter
+/// needs to be executed or it needs to be marked as `via_mutators`. Appart from that, they behave
+/// identically.
+///
 /// ```
 /// use typed_builder::TypedBuilder;
 ///
@@ -206,6 +214,13 @@ use core::ops::FnOnce;
 ///     }
 /// ))]
 /// struct Struct {
+///     // Does not require `self: (x)`, due to always
+//      // requiring, the field the mutator is specifed on.
+///     #[builder(mutators(
+///         fn x_into_b_field(self) {
+///             self.b.push(self.x)
+///         }
+///     ))]
 ///     x: i32,
 ///     #[builder(via_mutators(init = 1))]
 ///     a: i32,
@@ -215,8 +230,8 @@ use core::ops::FnOnce;
 ///
 /// // Mutators do not enforce only being called once
 /// assert_eq!(
-///     Struct::builder().x(2).x_into_b().x_into_b().inc_a(2).build(),
-///     Struct {x: 2, a: 3, b: vec![2, 2]});
+///     Struct::builder().x(2).x_into_b().x_into_b().x_into_b_field().inc_a(2).build(),
+///     Struct {x: 2, a: 3, b: vec![2, 2, 2]});
 /// ```
 pub use typed_builder_macro::TypedBuilder;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -121,6 +121,8 @@ use core::ops::FnOnce;
 ///    #[derive(Default)]
 ///    struct Point { x: f32, y: f32 }
 ///    ```
+/// - `mutators(...)` takes functions, that can mutate fields inside of the builder.
+///   See [mutators](#mutators) for details.
 ///
 /// On each **field**, the following values are permitted:
 ///
@@ -134,6 +136,12 @@ use core::ops::FnOnce;
 ///    derive proc-macro crates that complain about "expected literal".
 ///    Note that if `...` contains a string, you can use raw string literals to avoid escaping the
 ///    double quotes - e.g. `#[builder(default_code = r#""default text".to_owned()"#)]`.
+///
+/// - `via_mutators`: initialize the field when constructing the builder, useful in combination
+///   with [mutators](#mutators).
+///
+/// - `via_mutators = …` or `via_mutators(init = …)`: initialies the field with the expression `…`
+///   when constructing the builder, useful in combination with [mutators](#mutators).
 ///
 /// - `setter(...)`: settings for the field setters. The following values are permitted inside:
 ///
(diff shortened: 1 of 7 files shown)


typed-builder-macro/src/struct_info.rs
```diff
@@ -680,88 +680,70 @@ impl<'a> TypeBuilderAttr<'a> {
...
     }
 
-    fn apply_meta(&mut self, expr: syn::Expr) -> Result<(), Error> {
+    fn apply_meta(&mut self, expr: AttrArg) -> Result<(), Error> {
         match expr {
-            syn::Expr::Assign(assign) => {
-                let name =
-                    expr_to_single_string(&assign.left).ok_or_else(|| Error::new_spanned(&assign.left, "Expected identifier"))?;
-
+            AttrArg::KeyValue(ref kv @ KeyValue { ref name, ref value, .. }) => {
```
ModProg, line 685: I'm not sure, if the better approach might be to use `kv.name` and `kv.value` instead.
```suggestion
            AttrArg::KeyValue(kv) => {
```

typed-builder-macro/src/struct_info.rs
```diff
@@ -680,88 +680,70 @@ impl<'a> TypeBuilderAttr<'a> {
...
-                    "build_method" => {
-                        for arg in call.args {
-                            self.build_method.apply_meta(arg)?;
-                        }
-                        Ok(())
+                    Ok(())
+                }
+                "build_method" => {
+                    for arg in sub.args()? {
+                        self.build_method.apply_meta(arg)?;
```
ModProg, line 738: These currently fall back to using `syn::Expr` instead of `AttrArg`

typed-builder-macro/src/util.rs
```diff
@@ -125,3 +131,88 @@ pub fn expr_to_lit_string(expr: &syn::Expr) -> Result<String, Error> {
         _ => Err(Error::new_spanned(expr, "attribute only allows str values")),
     }
 }
+
+pub enum AttrArg {
+    Flag(Ident),
+    KeyValue(KeyValue),
+    Sub(SubAttr),
+    Not { not: Token![!], name: Ident },
+}
```
ModProg, line 140: Feel free to bikeshed, I'm also not sure about having `Not` as a struct variant instead of creating a `AttrNot{...}` struct.

ModProg:
I could also make `apply_subsections` generic and only use the `AttrArg` enum in the field attribute implementation.

typed-builder-macro/src/struct_info.rs
idanarye: Do you even _need_ `kv`? You are only using it in this line:

```rust
_ => Err(Error::new_spanned(kv, format!("Unknown parameter {:?}", name))),
```

But couldn't you just pass `expr`? It has the same tokens as `kv`.

typed-builder-macro/src/util.rs
idanarye: I do find value in consistency, but I don't mind struct variants. Even if they were all struct variants it'd be okay. That would require some other changes, which I might do after this PR is merged. So don't change it for now.

idanarye:
I've converted it to a draft. Change it back once it's ready.

typed-builder-macro/src/struct_info.rs
ModProg: korrect, but I'd still need to do `ref` on `name` and `value` anyway.

idanarye: Yes, you'd need that anyway, but that's a matter of style.

idanarye:
Mind if I manually cherry-pick 3f25b1588928ae3ca40a01a9087b7a5e89e520ee (and 28102d453095e279636d88d3bc8b9214d3bc86aa)? They are big changes that would conflict with any other change that wants to add new fields.

ModProg:
> Mind if I manually cherry-pick [3f25b15](https://github.com/idanarye/rust-typed-builder/commit/3f25b1588928ae3ca40a01a9087b7a5e89e520ee) (and [28102d4](https://github.com/idanarye/rust-typed-builder/commit/28102d453095e279636d88d3bc8b9214d3bc86aa))? They are big changes that would conflict with any other change that wants to add new fields.

feel free

idanarye:
#121 is merged.

ModProg:
rebase done

tests/tests.rs
```diff
@@ -801,3 +801,88 @@ fn test_issue_118() {
...
+        }
+        fn inc_x_by(self: (x), x: i32) {
+            self.x += x;
+        }
+        fn inc_preset(self) {
+            self.y += 1;
+            self.z += 1;
+            self.w += 1;
+        }
+        fn inc_y_by_x(self:(x)) {
```
idanarye, line 837: WTF? I thought the `: (x)` syntax means that only `self.x` is accessible - how come you can access `y` in this mutator?

Also - I'm not sure I like this syntax. It doesn't look like Rust. I think the proper way would be to use an attribute:

```rust
#[mutator(access(x, y))]
fn inc_y_by_x(self) {
```
Exact syntax and naming to be bikeshed.

Or maybe something like this:

```rust
fn inc_y_by_x(self: MutatorAccess!(x, y)) {
```

Where `MutatorAccess` will fake a struct that has x and y.

ModProg: > WTF? I thought the `: (x)` syntax means that only `self.x` is accessible - how come you can access `y` in this mutator?

As `via_mutators` are always available, I thought it wouldn't be necessary to specify them here. I.e. I had the (x) extend the Fields available.


> Also - I'm not sure I like this syntax. It doesn't look like Rust.

It kindof isn't, because it would mean self is a tuple struct, I can understand your hesitation here.

> I think the proper way would be to use an attribute:
> 
> ```rust
> #[mutator(access(x, y))]
> fn inc_y_by_x(self) {
> ```
> Exact syntax and naming to be bikeshed.
> 
> Or maybe something like this:
> 
> ```rust
> fn inc_y_by_x(self: MutatorAccess!(x, y)) {
> ```
> 
> Where `MutatorAccess` will fake a struct that has x and y.

I'm open to whatever you prefer here, wouldn't change much implementation wise.

ModProg: alright, will update. I think this will make the parsing probably cleaner even.

ModProg: > ```rust
> #[mutator(dependencies = [x])]
> fn inc_y_by_x(self) {
> ```

Naming wise, I also wondered about calling it `requires` or `required_fields`.

ModProg: Also, we could use `requires(x, y, z)` instead of `requires = [x, y, z]`. Not really a preference.

idanarye: I prefer `requires = [x, y, z]`. `requires(x, y, z)` makes it seem like `x`, `y` and `z` are sub-settings of `requires`, just like `requires` is a sub-setting of `mutator`.

src/lib.rs
```diff
@@ -9,7 +9,7 @@ use core::ops::FnOnce;
 /// Trying to set the same fields twice will generate a compile-time error. Trying to build without
 /// setting one of the fields will also generate a compile-time error - unless that field is marked
 /// as `#[builder(default)]`, in which case the `::default()` value of it's type will be picked. If
-/// you want to set a different default, use `#[builder(default=...)]`.
+/// you want to set a different default, use `#[builder(default=…)]`.
```
idanarye, line 12: Were these created automatically by something?

ModProg: oh, I think I did a search and replace

ModProg: I noticed that you used the unicode character and wanted to do that to what I added as well

idanarye: Didn't even notice it was in the rest of the docs until now. Turns out it was added in https://github.com/idanarye/rust-typed-builder/commit/3a748906c8f22ae0dfba74dcb8b24917a00da851.

After this gets in, I'm going to search-and-replace it with `...`. No reason to deviate from ASCII just for this...

typed-builder-macro/src/util.rs
```diff
@@ -289,4 +306,143 @@ pub trait ApplyMeta {
...
+    fn apply_attr(&mut self, attr: &Attribute) -> syn::Result<()> {
+        match &attr.meta {
+            syn::Meta::List(list) => self.apply_subsections(list),
+            meta => Err(Error::new_spanned(meta, "Expected builder(…)")),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Mutator {
```
idanarye, line 319: Nit: `Mutator` is not a utility, it can be in a new `mutator.rs` file.

idanarye (approved):
LGTM. There are some things I want to change, but this has been going on long enough (I failed to notice that you took it out of _draft_ two weeks ago) so I'm just going to merge it and maybe change them later.

typed-builder-macro/src/field_info.rs
```diff
@@ -243,6 +251,36 @@ impl ApplyMeta for FieldBuilderAttr<'_> {
...
+                        }
+                        if let Some(remaining) = args.next() {
+                            return Err(Error::new_spanned(remaining, "Expected only one argument (`init = ...`)"));
+                        }
+                        self.via_mutators = Some(value);
+                    }
+                }
+                Ok(())
+            }
+            "mutators" => expr.sub_attr()?.undelimited().map(|fns| self.mutators.extend(fns)),
```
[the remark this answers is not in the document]
idanarye: And now I see this has nothing to do with looping, and the map is related to error handling.

Issue #75, opened 2023-02-13
Context for builder


wackazong:
I would love to see a feature that allows to set a context for the builder. The context could then be used in the default functions. To keep the `builder()` function as is, there could be a `.context()` function to set the context. Then we could use something like::

`#[builder(default_code = "self.context()")]`

What do you think?

I find the code of this crate quite complex for me as a Rust beginner, but I would be happy to do a PR if given some pointers on where to start.

wackazong:
My use case would be to allow for the control of default values for keys and timestamps in different testing environments.

idanarye:
> To keep the `builder()` function as is, there could be a `.context()` function to set the context.

So... something like this?

```rust
#[derive(TypedBuilder)]
#[builder(context(/*Syntax TBD*/))]
struct Foo {
    #[buidler(default = self.context())]
    bar: i32,
}

// This will work:
Foo::builder().context(Context::new()).build();

// This will work:
Foo::builder().bar(42).build();

// This will not:
Foo::builder().build();
```

This can be a bit hard to model, but what if we require the context to always have a default?

Also, rather than a "context" method, I'd rather do something ike this:

```rust
#[derive(TypedBuilder)]
#[builder(context(
    // Will be treated like a field that does not get into the final struct
    number: i32 = 4,
))]
struct Foo {
    #[buidler(default = number * 10 + 2)]
    bar: i32,
}
```

wackazong:
Yes, the first proposition would be great. The context would just need to be available in the builder, not in the final struct. I would like to set the context at runtime, therefore an attribute macro as you suggested would not serve me. 

Maybe like this?

```
struct Context;

impl Context {
    fn bar() {
        ...
    }
}

#[derive(TypedBuilder)]
struct Foo {
    #[builder(use_context_for_default)]
    bar: i32,
}

// if a field is marked with use_context_for_default then a context needs to be supplied
// and the default value is derived from the context passed in at runtime

// context will be required for each call of builder if default value is not given
Foo::builder().context(Context::new()).build();

// This will work:
Foo::builder().bar(42).build();

// This will not:
Foo::builder().build();
```

idanarye:
I really don't want to move the defaults elsewhere. If I have to specify dependencies, I'd rather make a breaking change and do something like this:

```rust
#[derive(TypedBuilder)]
#[builder(context(
    x: i32 = 1,
    y: i32 = 2,
))]
struct MyStruct {
    // Depends only on `x`
[...]

PR #120, opened 2023-09-22
Mutable assignments variables

novafacing:
Very small change to allow a use case for LibAFL where a `default = ` expression needs to mutate another field. Here's a snippet for example:

```rust
/// Sending end on a (unidirectional) sharedmap channel
#[derive(TypedBuilder, Debug)]
#[builder(build_method(into = Result::<LlmpSender<SP>, Error>))]
pub struct LlmpSender<SP>
where
    SP: ShMemProvider,
{
    /// ID of this sender.
    id: ClientId,
    #[builder(default = ptr::null_mut())]
    /// Ref to the last message this sender sent on the last page.
    /// If null, a new page (just) started.
    last_msg_sent: *const LlmpMsg,
    #[builder(default)]
    /// A vec of pages that we previously used, but that have served its purpose
    /// (no potential readers are left).
    /// Instead of freeing them, we keep them around to potentially reuse them later,
    /// if they are still large enough.
    /// This way, the OS doesn't have to spend time zeroing pages, and getting rid of our old pages
    unused_shmem_cache: Vec<LlmpSharedMap<SP::ShMem>>,
    #[builder(default = false)]
    /// If true, pages will never be pruned.
    /// The broker uses this feature.
    /// By keeping the message history around,
    /// new clients may join at any time in the future.
    keep_pages_forever: bool,
    #[builder(default = false)]
    /// True, if we allocatd a message, but didn't call [`Self::send()`] yet
    has_unsent_message: bool,
    /// The sharedmem provider to get new sharaed maps if we're full
    shmem_provider: SP,
    #[builder(
        default = vec![LlmpSharedMap::new(id, shmem_provider.new_shmem(LLMP_CFG_INITIAL_MAP_SIZE).expect("Failed to create a new shared memory"))],
        setter(into))
    ]
    /// A vec of page wrappers, each containing an initialized [`ShMem`]
    out_shmems: Vec<LlmpSharedMap<SP::ShMem>>,
}
```

### DIFF
--- a/typed-builder-macro/src/field_info.rs
+++ b/typed-builder-macro/src/field_info.rs
@@ -116,6 +116,7 @@ pub struct FieldBuilderAttr<'a> {
     pub default: Option<syn::Expr>,
     pub deprecated: Option<&'a syn::Attribute>,
     pub setter: SetterSettings,
+    pub mutable_during_default_resolution: bool,
 }
 
 #[derive(Debug, Default, Clone)]
@@ -200,6 +201,10 @@ impl<'a> FieldBuilderAttr<'a> {
                 match name.as_str() {
                     "default" => {
                         self.default = Some(syn::parse2(quote!(::core::default::Default::default())).unwrap());
+                        Ok(())
+                    }
+                    "mutable_during_default_resolution" => {
+                        self.mutable_during_default_resolution = true;
                         Ok(())
                     }
                     _ => Err(Error::new_spanned(&path, format!("Unknown parameter {:?}", name))),
--- a/typed-builder-macro/src/field_info.rs
+++ b/typed-builder-macro/src/field_info.rs
@@ -115,7 +115,7 @@ pub struct FieldBuilderAttr<'a> {
     pub deprecated: Option<&'a syn::Attribute>,
     pub setter: SetterSettings,
     pub mutators: Vec<ItemFn>,
-    pub mutable_during_default_resolution: bool,
+    pub mutable_during_default_resolution: Option<Span>,
 }
 
 #[derive(Debug, Default, Clone)]
@@ -239,10 +239,10 @@ impl ApplyMeta for FieldBuilderAttr<'_> {
                 Ok(())
             }
             "setter" => self.setter.apply_sub_attr(expr),
-            "mutable_during_default_resolution" => {
-                self.mutable_during_default_resolution = true;
-                Ok(())
-            }
+            "mutable_during_default_resolution" => expr.apply_flag_to_field(
+                &mut self.mutable_during_default_resolution,
+                "made mutable during default resolution",
+            ),
             _ => Err(Error::new_spanned(
                 expr.name(),
                 format!("Unknown parameter {:?}", expr.name().to_string()),
--- a/typed-builder-macro/src/struct_info.rs
+++ b/typed-builder-macro/src/struct_info.rs
@@ -500,7 +500,7 @@ impl<'a> StructInfo<'a> {
         };
 
         quote!(
-            #[allow(dead_code, non_camel_case_types, missing_docs)]
+            #[allow(dead_code, non_camel_case_types, missing_docs, unused_mut)]
             #[automatically_derived]
             impl #impl_generics #builder_name #modified_ty_generics #where_clause {
                 #build_method_doc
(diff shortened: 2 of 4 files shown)


idanarye:
I'm... not 100% sure how I feel about this. I'm still on the fence about allowing `default` expressions access to the previous fields - this is an undocumented feature, which may prevent the ability to split things out to traits in the future.

I'll allow it, but keep in mind that it might be broken in the future. Probably in favor of something like #75 (which in this case would mean the `shmem_provider` would be passed as a context, and then maybe moved/cloned into its own field)

One change I must insist on though - one of Rust's basic tenants is that immutable should be the default, and I really don't think this rule should be violated here. The mutability should be triggered by something in the attribute:

```rust
/// The sharedmem provider to get new sharaed maps if we're full
#[builder(mutable_during_default_resolution)]
shmem_provider: SP,
```

Yes, it's long. Ugly features get ugly names.

Also, this means it needs to wait until #116 gets in. Or at least until [I cherry-pick some commits from it](https://github.com/idanarye/rust-typed-builder/pull/116#issuecomment-1734481757).

novafacing:
Sure thing and sound reasoning, I'll make that change! I'll stay in tune with updates as they happen and in progress PRs. Happy to work on making mutability more idiomatic in the future  -- for now, this will be a very helpful stopgap!

idanarye:
#121 is in. It should now be easier to add new fields.

novafacing:
Ok, added a doc entry and added that field attribute!

typed-builder-macro/src/struct_info.rs
```diff
@@ -462,10 +462,30 @@ impl<'a> StructInfo<'a> {
             let name = &field.name;
             if let Some(ref default) = field.builder_attr.default {
                 if field.builder_attr.setter.skip.is_some() {
-                    quote!(let #name = #default;)
+                    if field.builder_attr.mutable_during_default_resolution {
```
idanarye, line 465: Instead of adding the mutability each time, you can add it afterwards:

```rust
let field_assignment = if let Some(ref default) = field.builder_attr.default {
    // Original code
} else {
    // This is the same as the original:
     quote!(let #name = #name.0;)
};
if field.builder_attr.mutable_during_default_resolution {
    quote! {
        #field_assignment
        let mut #name = #name;
    }
} else {
    field_assignment
}
```

Or maybe:
```rust
let field_value = if let Some(ref default) = field.builder_attr.default {
    // Original code, but without the `let #name =` and the `;` in the quotes
} else {
    // Like this:
     quote!(#name.0)
};
if field.builder_attr.mutable_during_default_resolution {
    quote!(let mut #name = #field_value;)
} else {
    quote!(let #name = #field_value;)
}
```

Or even better:

```rust
let maybe_mut = if field.builder_attr.mutable_during_default_resolution {
    quote!(mut)
} else {
    quote!()
};
if let Some(ref default) = field.builder_attr.default {
    // Original code, but use `#maybe_mut` to set the mutability
} else {
    // Like this:
     quote!(let #maybe_mut #name = #name.0;)
}
```

I think this method is preferable because we can then later add a span to `maybe_mut` and get better warnings.

typed-builder-macro/src/field_info.rs
```diff
@@ -238,6 +239,10 @@ impl ApplyMeta for FieldBuilderAttr<'_> {
                 Ok(())
             }
             "setter" => self.setter.apply_sub_attr(expr),
+            "mutable_during_default_resolution" => {
+                self.mutable_during_default_resolution = true;
```
idanarye, line 243: Please add this line above:
```rust
expr.flag()?
```
To enforce that `mutable_during_default_resolution` is used correctly.

Even better - if you change the field type to `Span<Option>` you could use this:
```rust
"mutable_during_default_resolution" => expr.apply_flag_to_field(
    &mut self.mutable_during_default_resolution,
    "made mutable during default resolution",
),
```

Which would allow setting `mutable_during_default_resolution` in field defaults and removing it with `!mutable_during_default_resolution` for individual fields. The span is needed for better error reporting - and if we add it to the `mut` we'll get better "unused `mut`" warnings.

idanarye (changes requested):
If any of these remarks are too difficult, I can do them myself after I merge.

idanarye (approved):

PR #121, opened 2023-09-26
Change `AttrArg` workflow - match on name first, type second

idanarye:


idanarye:
@ModProg - I've taken your `AttrArg` idea and modified it so that the name will be matched first. This means that if a setting is used with the wrong style, instead of getting an error that it does not exist we'd get an error that it is of the wrong type. Currently this error does not specify what the correct type is, but that can be added in the future.

idanarye:
I'm also considering adding helper functions like `iter_if_sub` for the other style, but I'm not 100% sure yet if `iter_if_sub` is the correct abstraction...

idanarye:
My `iter_if_sub` is similar to your `try_get_subcommand`, but now that I think about it,
```rust
for arg in expr.iter_if_sub()? {
```
is not that much simpler over
```rust
for arg in expr.subcommand()?.args()? {
```

So maybe I should just have `flag()`, `keyvalue()`, and `subcommand()` methods? The naming convention comes from how `Result` has `ok()` and `err()` methods. The semantics are a bit different (they return `Some` while I return `Result`, but the idea is similar.

I also want to have `_or_nullify()` version of all these methods, which will accept a `&mut Option<T>` and handle `AttrArg::Not`.

idanarye:
Actually, `_or_nullify` is a problem because it must still return a value (which does not exist) or an error. Maybe `or_not()`, which would return and `Option`?

ModProg:
> Actually, `_or_nullify` is a problem because it must still return a value (which does not exist) or an error. Maybe `or_not()`, which would return and `Option`?

that makes more sense, I feel apis taking &mut to be a bit clunky

Issue #84, opened 2023-05-26
`const fn` doesn't work


StackOverflowExcept1on:
```rust
use typed_builder::TypedBuilder;

#[derive(TypedBuilder)]
struct Foo {
    // Mandatory Field:
    x: i32,

    // #[builder(default)] without parameter - use the type's default
    // #[builder(setter(strip_option))] - wrap the setter argument with `Some(...)`
    #[builder(default, setter(strip_option))]
    y: Option<i32>,

    // Or you can set the default
    #[builder(default=20)]
    z: i32,
}

fn f() {
    const X: Foo = Foo::builder().z(1).x(2).y(3).build();
}

fn main() {}
```

```
error[E0015]: cannot call non-const fn `Foo::builder` in constants
  --> src/main.rs:19:20
   |
19 |     const X: Foo = Foo::builder().z(1).x(2).y(3).build();
   |                    ^^^^^^^^^^^^^^
   |
   = note: calls in constants are limited to constant functions, tuple structs and tuple variants

error[E0015]: cannot call non-const fn `FooBuilder::z` in constants
  --> src/main.rs:19:35
   |
19 |     const X: Foo = Foo::builder().z(1).x(2).y(3).build();
   |                                   ^^^^
   |
   = note: calls in constants are limited to constant functions, tuple structs and tuple variants
```

idanarye:
I can try adding a top-level `#[builder(const)]` attribute to make all the generated functions `const`. Keep in mind though that with the current version of Rust this example could never work, because `Default::default` is not `const`. So having a `const` builder would forbid bare `#[builder(default)]`s - the one for `y` will have to change to `#[builder(default = None)]`.

StackOverflowExcept1on:
also there are some problems with `const fn`:
```rust
struct Foo {
    s: Option<String>,
}

impl Foo {
    const fn new() -> Self {
        Self { s: None }
    }

    const fn s(mut self, s: String) -> Self {
        self.s = Some(s); //error[E0493]: destructor of `Option<String>` cannot be evaluated at compile-time
        self
    }
}
```

idanarye:
I don't think that'd be a problem. Typed builder's setters do not override fields, so they don't need to invoke destructors.

idanarye:
Due to the way I parse things, it's going to have to be `#[builder(const_)]` and not `#[builder(const)]`.

idanarye:
[...]

PR #123, opened 2023-10-15
Code cleanup. Make build method generate as const

DanikVitek:
- Add `#![forbid(rust_2018_idioms)]` to the `typed-builder-macro` crate
- Make the builder function const when possible

### DIFF
--- a/typed-builder-macro/src/mutator.rs
+++ b/typed-builder-macro/src/mutator.rs
@@ -52,7 +52,7 @@ impl ApplyMeta for MutatorAttribute {
 }
 
 impl Parse for Mutator {
-    fn parse(input: ParseStream) -> syn::Result<Self> {
+    fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
         let mut fun: ItemFn = input.parse()?;
 
         let mut attribute = MutatorAttribute::default();
--- a/typed-builder-macro/src/struct_info.rs
+++ b/typed-builder-macro/src/struct_info.rs
@@ -1,8 +1,7 @@
 use proc_macro2::{Ident, Span, TokenStream};
 use quote::{format_ident, quote, quote_spanned, ToTokens};
-use syn::parse::Error;
 use syn::punctuated::Punctuated;
-use syn::{parse_quote, GenericArgument, ItemFn, Token};
+use syn::{parse_quote, Error, GenericArgument, ItemFn, Token};
 
 use crate::field_info::{FieldBuilderAttr, FieldInfo};
 use crate::mutator::Mutator;
@@ -49,7 +48,7 @@ impl<'a> StructInfo<'a> {
             .collect()
     }
 
-    pub fn new(ast: &'a syn::DeriveInput, fields: impl Iterator<Item = &'a syn::Field>) -> Result<StructInfo<'a>, Error> {
+    pub fn new(ast: &'a syn::DeriveInput, fields: impl Iterator<Item = &'a syn::Field>) -> syn::Result<Self> {
         let builder_attr = TypeBuilderAttr::new(&ast.attrs)?;
         let builder_name = builder_attr
             .builder_type
@@ -70,7 +69,7 @@ impl<'a> StructInfo<'a> {
         })
     }
 
-    pub fn builder_creation_impl(&self) -> Result<TokenStream, Error> {
+    pub fn builder_creation_impl(&self) -> syn::Result<TokenStream> {
         let StructInfo {
             vis,
             ref name,
@@ -177,10 +176,10 @@ impl<'a> StructInfo<'a> {
             impl #impl_generics #name #ty_generics #where_clause {
                 #builder_method_doc
                 #[allow(dead_code, clippy::default_trait_access)]
-                #builder_method_visibility fn #builder_method_name() -> #builder_name #generics_with_empty {
+                #builder_method_visibility const fn #builder_method_name() -> #builder_name #generics_with_empty {
                     #builder_name {
                         fields: (#(#init_fields_expr,)*),
-                        phantom: ::core::default::Default::default(),
+                        phantom: ::core::marker::PhantomData,
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


typed-builder-macro/src/mutator.rs
```diff
@@ -52,7 +52,7 @@ impl ApplyMeta for MutatorAttribute {
 }
 
 impl Parse for Mutator {
-    fn parse(input: ParseStream) -> syn::Result<Self> {
+    fn parse(input: ParseStream<'_>) -> syn::Result<Self> {
```
idanarye, line 55: This seems redundant. Is this because of that `#![forbid(rust_2018_idioms)]` thing?

typed-builder-macro/src/struct_info.rs
```diff
@@ -16,7 +18,7 @@ pub struct StructInfo<'a> {
     pub vis: &'a syn::Visibility,
     pub name: &'a syn::Ident,
     pub generics: &'a syn::Generics,
-    pub fields: Vec<FieldInfo<'a>>,
+    pub fields: Box<[FieldInfo<'a>]>,
```
idanarye, line 21: Was this really necessary?

typed-builder-macro/src/struct_info.rs
```diff
@@ -85,15 +87,27 @@ impl<'a> StructInfo<'a> {
...
-        let init_fields_expr = self.included_fields().map(|f| {
-            f.builder_attr.via_mutators.as_ref().map_or_else(
-                || quote!(()),
-                |via_mutators| {
-                    let init = &via_mutators.init;
-                    quote!((#init,))
-                },
-            )
-        });
+        let builder_method_const = Rc::new(OnceCell::new());
```
idanarye, line 90: Respectfully WTF? Looking at the code, it seems like a simple:
```rust
let mut builder_method_const = true;
```
would suffice.

typed-builder-macro/src/struct_info.rs
```diff
@@ -85,15 +87,27 @@ impl<'a> StructInfo<'a> {
...
+        let init_fields_expr = self
+            .included_fields()
+            .map({
+                let builder_method_const = builder_method_const.clone();
+                move |f| {
+                    f.builder_attr.via_mutators.as_ref().map_or_else(
```
idanarye, line 96: Does no one use simple `if let`s nowadays?

typed-builder-macro/src/struct_info.rs
```diff
@@ -85,15 +87,27 @@ impl<'a> StructInfo<'a> {
...
+                        || quote!(()),
+                        |via_mutators| {
+                            let init = &via_mutators.init;
+                            if !matches!(init, syn::Expr::Lit(_)) {
```
idanarye, line 100: I like to make things explicit. If the build method is to be `const`, it should be marked as such in the configuration attributes - not implicitly deduced based on some incomplete (literals aren't the only `const` expressions) heuristics.

idanarye:
Given how the setter methods are not `const`, and the `build` method is not `const`, is it really useful to make the `builder` method `const`?

typed-builder-macro/src/struct_info.rs
```diff
@@ -70,7 +72,7 @@ impl<'a> StructInfo<'a> {
         })
     }
 
-    pub fn builder_creation_impl(&self) -> Result<TokenStream, Error> {
+    pub fn builder_creation_impl(&self) -> syn::Result<TokenStream> {
```
idanarye, line 75: TBH this is the only change I like in this PR.

typed-builder-macro/src/struct_info.rs
DanikVitek: Borrow checker was not happy about the moved value. It could be done with `Rc<Cell<bool>>` though.

typed-builder-macro/src/mutator.rs
DanikVitek: Yes. This lets to see at a glance where a borrowing occurs

typed-builder-macro/src/struct_info.rs
DanikVitek: Well, it is not very usefull, but the field is effectively immutable and so there is no need to take away additional 8 bytes for capacity. But the struct is already >2KiB in size so tbh no much difference

typed-builder-macro/src/struct_info.rs
DanikVitek: True, makes sense. It was just my compulsive thought to make it const when I saw the macro expansion for my usecase. But now I know that it is not so simple

typed-builder-macro/src/struct_info.rs
DanikVitek: If you're about the `matches!` macro, then there would be a dangling branch, because it needs an inverted pattern — not `Lit(_)`

DanikVitek:
> Given how the setter methods are not `const`, and the `build` method is not `const`, is it really useful to make the `builder` method `const`?

Not really. I'm not sure if it would be possible because of deconstruction of `self`. Currently it's not possible in const fn afaik

typed-builder-macro/src/struct_info.rs
idanarye: I tried it locally and it worked. Had to remove the `move` from the closure in the `map`, of course.

I speculate that it didn't work for you because you tried it before adding the `collect`. The compiler, in a way, was complaining about the very same bug that the `collect` solved.

Another option is to just do a second iteration:

```rust
let builder_method_const = self
    .included_fields()
    .filter_map(|field| field.builder_attr.via_mutators.as_ref())
    .all(|via_mutators| matches!(via_mutators.init, syn::Expr::Lit(_)))
    .then(|| quote!(const));
```

But of course - all of this is irrelevant because https://github.com/idanarye/rust-typed-builder/pull/123#discussion_r1359976876.

typed-builder-macro/src/struct_info.rs
idanarye: No, it's about the `map_or_else`. But I just noticed that it was added by someone else, so nevermind.

idanarye:
Once https://github.com/rust-lang/rust/issues/67792 gets stabilized, `builder`, `build`, and all the setters could be `const` - and that would be controlled via a flag on the attribute (#84). Until then, I see no point in making only the `builder` method `const`.

PR #125, opened 2023-10-18
Add support for `?Sized` generics

aumetra:
This PR adds support for structs with `?Sized` generic parameters. Previously those were not accepted.

Technically this is a breaking change since this changes the internal `PhantomData` type from `PhantomData<(T1, T2)>` to `(PhantomData<T1>, PhantomData<T2>)`, and `Default` is only implemented for tuples of up to 12 items due to missing variadic generics.

Practically no struct should hit this limitation though.

### DIFF
--- a/typed-builder-macro/src/struct_info.rs
+++ b/typed-builder-macro/src/struct_info.rs
@@ -188,9 +188,9 @@ impl<'a> StructInfo<'a> {
             #[must_use]
             #builder_type_doc
             #[allow(dead_code, non_camel_case_types, non_snake_case)]
-            #builder_type_visibility struct #builder_name #b_generics {
+            #builder_type_visibility struct #builder_name #b_generics #b_generics_where_extras_predicates {
                 fields: #all_fields_param,
-                phantom: ::core::marker::PhantomData<(#( #phantom_generics ),*)>,
+                phantom: (#( ::core::marker::PhantomData<#phantom_generics> ),*),
             }
 
             #[automatically_derived]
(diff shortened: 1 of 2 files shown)


idanarye:
> Practically no struct should hit this limitation though.

Famous last words?

Why not make it `PhantomData<(PhantomData<T1>, PhantomData<T2>)>`? That way the outer phantom data will not be a tuple and will have a `Default` no matter how many generic arguments the type will have.

aumetra:
> Famous last words?

Heh, fair

> Why not make it `PhantomData<(PhantomData<T1>, PhantomData<T2>)>`?

I honestly just didn't think about it. Changed it

typed-builder-macro/src/struct_info.rs
```diff
@@ -188,9 +188,9 @@ impl<'a> StructInfo<'a> {
             #[must_use]
             #builder_type_doc
             #[allow(dead_code, non_camel_case_types, non_snake_case)]
-            #builder_type_visibility struct #builder_name #b_generics {
+            #builder_type_visibility struct #builder_name #b_generics #b_generics_where_extras_predicates {
```
idanarye, line 191: BTW - why is this needed? `b_generics_where_extras_predicates` should already be contained inside `b_generics`.

aumetra: I had to add this because it didn't add the `where T: ?Sized` clause.  
Looking at the [`ToTokens` impl](https://github.com/dtolnay/syn/blob/e20379473b266ec5904bc87415d80de3ab1ac059/src/generics.rs#L980-L1013) of `Generics`, the `where` clause is absent, so it has to be added manually.

idanarye: This is certainly weird. I'll look into it locally after merging.

PR #128, opened 2023-10-31
Use format_ident! macro to build formatted identifiers

itchyny:
The [`format_ident!`](https://docs.rs/quote/latest/quote/macro.format_ident.html) macro saves us the code for stripping `r#`.


idanarye:
I've mereged #127, so please rebase this PR.

PR #129, opened 2023-11-12
Support deriving builder for enums

itchyny:
The name of builder methods are the "snake_case" of variant names. Note that it does not support enums with generics or lifetime.


itchyny:
Gentle ping, ayn thoughts on enum support?

typed-builder-macro/src/enum_info.rs
```diff
@@ -0,0 +1,158 @@
...
+        if builder_attr.builder_type.name.is_some() {
+            return Err(Error::new_spanned(
+                ast,
+                "TypedBuilder is not supported for enum with builder_type(name=...)",
+            ));
+        }
+        if !matches!(builder_attr.build_method.into, IntoSetting::NoConversion) {
+            return Err(Error::new_spanned(
+                ast,
+                "TypedBuilder is not supported for enum with build_method(into=...)",
```
idanarye, line 38: Ditto

typed-builder-macro/src/enum_info.rs
```diff
@@ -0,0 +1,158 @@
+use convert_case::{Case, Casing};
+use proc_macro2::{Span, TokenStream};
+use quote::{format_ident, quote, ToTokens};
+use syn::{parse::Error, parse_quote, punctuated::Punctuated, Token};
+
+use crate::builder_attr::{IntoSetting, TypeBuilderAttr};
+use crate::struct_info::StructInfo;
+
+pub struct EnumInfo<'a> {
```
idanarye, line 9: I don't like this struct. Unlike `StructInfo` and `FieldInfo`, which hold a parsed representation of the struct/field ready to create the builder code from, this `EnumInfo` here holds direct references to the AST. It's `new` method does no parsing - only verification.

typed-builder-macro/src/enum_info.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    fn derive_variant_impl(
+        &self,
+        variant_name: &syn::Ident,
+        variant_attrs: &[syn::Attribute],
+        variant_fields: &syn::FieldsNamed,
+    ) -> syn::Result<TokenStream> {
+        let enum_name = &self.ast.ident;
+        let internal_struct_name = format_ident!("{}{}", enum_name, variant_name);
+        let internal_struct_ast = &syn::DeriveInput {
```
idanarye, line 55: I'd rather create a `StructInfo` from the variant than fabricate a `DeriveInput`. Even if it means altering `StructInfo::new` to support this.

idanarye: Creating a `StructInfo` using the constructor is also acceptable here - most of the fields you need to fill are ones you manually fabricate in this `DeriveInput` anyway.

typed-builder-macro/src/enum_info.rs
```diff
@@ -0,0 +1,158 @@
...
+            impl #enum_name {
+                #[doc = #builder_method_doc]
+                #[allow(dead_code)]
+                #builder_method_visibility fn #builder_method_name() -> #internal_builder_name {
+                    #internal_struct_name::#internal_builder_method_name()
+                }
+            }
+
+            #[automatically_derived]
+            impl From<#internal_struct_name> for #enum_name {
```
idanarye, line 134: I don't like that we define a potentially visible internal struct just for this. I'd rather modify `DeriveInput` to have an `Option`al "parent enum" field, and have it directly generate the enum variant of that field.

idanarye (commented):
Right, sorry about that. I've actually started reviewing a few weeks ago, but things happened and I got distracted and left it half-reviewed...

PR #136, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    open-pull-requests-limit: 10
+  - package-ecosystem: "docker"
```
idanarye, line 8: Is this necessary? I'm not using Docker here. Unless GitHub Actions uses it behind the scenes - but wouldn't that be handled by the `package-ecosystem: "github-actions"` section?

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+    directory: "/"
+    schedule:
+      interval: "monthly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
```
idanarye, line 13: I'm not familiar with dependabot (did not even know about its existence) but isn't the directory for GitHub Actions `.github/workflows`?

.github/dependabot.yml
oriontvv: Ok, I added common template - will remove docker section. Fyi, if you don't have any Dockerfiles - the job would just skip it

.github/dependabot.yml
oriontvv: Not really, it's the correct path

.github/dependabot.yml
idanarye: If it's a common template then leave it as is. No reason to deviate from the standards.

PR #141, opened 2024-04-15
Include license files in typed-builder-macro crate

decathorpe:
Both the Apache-2.0 and MIT licenses require that (re)distributed sources contain a copy of the original license text. As of right now, the sources published and redistributed via crates.io for the typed-builder-macro crate do not contain license texts.

This PR adds symbolic links that point at the license texts in the project's root directory, which is enough for "cargo package" / "cargo publish" to resolve and include them when publishing to crates.io in most circumstances (unless you're running the "cargo publish" command on Windows, AFAIK).

### DIFF
--- a/typed-builder-macro/LICENSE-APACHE
+++ b/typed-builder-macro/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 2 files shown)


idanarye:
Any reason not to just copy the files? It's not like they are going to change...

decathorpe:
> Any reason not to just copy the files? It's not like they are going to change...

Not really. Looking at other projects where multiple crates are published from the same repo, using symbolic links just seems to be the standard practice.

PR #146, opened 2024-06-15
Apply field doc-comments to builder methods

sunsided:
I noticed that the builder has support for explicitly provided comments, e.g. via

```rust
#[builder(
    default,
    setter(strip_option, doc = "Set `y`. If you don't specify a value it'll default to no value.",)
)]
```

However, regular doccomments would be ignored even when no setter-specific `doc` attribute is provided. With this PR, the following is now possible:

```rust
struct Foo {
    /// `x` value.
    ///
    /// This field is mandatory.
    x: i32,

    /// This doccomment is ignored for the setter. 
    #[builder(
        default,
        setter(strip_option, doc = "Set `y`. If you don't specify a value it'll default to no value.",)
    )]
    y: Option<i32>,

    #[builder(default = 20)]
    z: i32,
}
```

On the builder, `x(...)` will be documented with the regular doccomment, `y(...)` is documented with the setter attribute as before and `z(...)` remains undocumented, as before.

When the builder emits `deprecated` functions, they also inherit the documentation. This was not previously the case.

### DIFF
--- a/examples/example.rs
+++ b/examples/example.rs
@@ -2,12 +2,17 @@ use typed_builder::TypedBuilder;
 
 #[derive(Debug, PartialEq, TypedBuilder)]
 struct Foo {
-    // Mandatory Field:
+    /// `x` value.
+    ///
+    /// This field is mandatory.
     x: i32,
 
     // #[builder(default)] without parameter - use the type's default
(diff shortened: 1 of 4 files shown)


idanarye:
Thanks!

PR #150, opened 2024-08-21
feat: add `strip_option(fallback = field_opt)`

### LINKED ISSUE #117: feature: add an `opt` version of the builders when `strip_option` is enabled.
Hi!

Currently `strip_option` prevents the actual setting of `None`, forcing to use  two code paths when one wants to maybe build with `None`.
I propose the following. When using `strip_option`, a second setter is added, with an `_opt` suffix, that takes an option.

Here is an example:

### Current version

```rust
#[derive(TypedBuilder)]
struct SomeStruct {
    #[builder(default, setter(strip_option)]
     inner: Option<u32>,
}

fn some_fn(value: Option<u32>) -> SomeStruct {
    match value {
        None => SomeStruct::builder().build(),
        Some(v) => SomeStruct::builder().inner(v).build(),
    }
}
```

### Proposal:

```rust
#[derive(TypedBuilder)]
struct SomeStruct {
    #[builder(default, setter(strip_option)]
     inner: Option<u32>,
}

/// The old way still works the same
fn old_way(value: Option<u32>) -> SomeStruct {
    match value {
        None => SomeStruct::builder().build(),
        Some(v) => SomeStruct::builder().inner(v).build(),
    }
}

fn new_way(value: Option<u32>) -> SomeStruct {
    /// There are now 2 ways to set `inner`, allowing both uses, but privileging the `Some` case as intended.
    SomeStruct::builder().inner_opt(value).build()
}
```

### Limitations

This would be a breaking change, since it would break the compilation of structures having for example an `inner` and an `inner_opt` field. This could be made opt-in for example by requiring the user to give a name to the `_opt` version.

ifiokjr:
### Description

Fixes #117 using the API described in https://github.com/idanarye/rust-typed-builder/issues/117#issuecomment-2178009713.

```rust
#[derive(TypedBuilder)]
struct SomeStruct {
     #[builder(default, setter(strip_option)]
     inner: Option<u32>,
     #[builder(setter(strip_option(fallback="outer_opt"))]
     outer: Option<u32>,

}

/// `strip_option` retains its current behaviour avoiding a breaking change.
/// `strip_option(fallback ="outer_opt")` strips the `Option`, and creates a fallback that still takes the option. 
/// This makes it fully opt-in.
fn new_api(value: Option<u32>) -> SomeStruct {
    match value {
        None => SomeStruct::builder().outer_opt(None).build(),
        Some(v) => SomeStruct::builder().inner(v).outer(v).build(),
    }
}

/// `strip_option(fallback ="outer_opt") makes the original `outer` builder method take the value, and creates an `outer_opt` which still takes an `Option`.
fn new_way(value: Option<u32>) -> SomeStruct {
    SomeStruct::builder().outer_opt(value).build()
}
```

### DIFF
--- a/typed-builder-macro/src/field_info.rs
+++ b/typed-builder-macro/src/field_info.rs
@@ -3,9 +3,7 @@ use quote::quote_spanned;
 use syn::{parse::Error, spanned::Spanned};
 
 use crate::mutator::Mutator;
-use crate::util::{
-    expr_to_lit_string, ident_to_type, path_to_single_string, strip_raw_ident_prefix, ApplyMeta, AttrArg, SubAttr,
-};
+use crate::util::{expr_to_lit_string, ident_to_type, path_to_single_string, strip_raw_ident_prefix, ApplyMeta, AttrArg};
 
 #[derive(Debug)]
 pub struct FieldInfo<'a> {
@@ -129,7 +127,7 @@ pub struct SetterSettings {
     pub doc: Option<syn::Expr>,
     pub skip: Option<Span>,
     pub auto_into: Option<Span>,
-    pub strip_option: Option<StripOption>,
+    pub strip_option: Option<Strip>,
     pub strip_bool: Option<Span>,
     pub transform: Option<Transform>,
     pub prefix: Option<String>,
@@ -197,7 +195,7 @@ impl<'a> FieldBuilderAttr<'a> {
 
         let conflicting_transformations = [
             ("transform", self.setter.transform.as_ref().map(|t| &t.span)),
-            ("strip_option", self.setter.strip_option.as_ref().map(|s| s.span())),
+            ("strip_option", self.setter.strip_option.as_ref().map(|s| &s.span)),
             ("strip_bool", self.setter.strip_bool.as_ref()),
         ];
         let mut conflicting_transformations = conflicting_transformations
@@ -339,21 +337,23 @@ impl ApplyMeta for SetterSettings {
             "into" => expr.apply_flag_to_field(&mut self.auto_into, "calling into() on the argument"),
             "strip_option" => {
                 let caption = "putting the argument in Some(...)";
+
                 match expr {
                     AttrArg::Sub(sub) => {
+                        let span = sub.span();
+                        let mut strip_option = Strip::new(span);
+
                         if self.strip_option.is_none() {
-                            self.strip_option = Some(StripOption::from_args(sub)?);
+                            strip_option.apply_sub_attr(sub)?;
+                            self.strip_option = Some(strip_option);
                             Ok(())
                         } else {
-                            Err(Error::new(
-                                sub.span(),
-                                format!("Illegal setting - field is already {caption}"),
-                            ))
+                            Err(Error::new(span, format!("Illegal setting - field is already {caption}")))
                         }
                     }
                     AttrArg::Flag(flag) => {
                         if self.strip_option.is_none() {
-                            self.strip_option = Some(StripOption::Span(flag.span()));
+                            self.strip_option = Some(Strip::new(flag.span()));
                             Ok(())
                         } else {
                             Err(Error::new(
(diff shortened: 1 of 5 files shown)


ifiokjr:
@idanarye maybe I can extend the functionality here to also cover `strip_bool`.

typed-builder-macro/src/field_info.rs
```diff
@@ -345,6 +378,58 @@ impl ApplyMeta for SetterSettings {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum StripOption {
```
idanarye, line 382: Nit: a better design would have been:

```rust
pub struct StripOption {
    pub fallback: Option<String>, // not sure it should even be a string, but that's for another comment
    span: Span,
}
```

That would allow for adding more sub-settings in the future, and also make it possible to use `ApplyMeta` for parsing `StripOption`.

typed-builder-macro/src/field_info.rs
```diff
@@ -345,6 +378,58 @@ impl ApplyMeta for SetterSettings {
...
+    Span(Span),
+    WithFallback(Span, String),
```
idanarye, line 384: Rather than a string, why not use `syn::Ident`? This would also mean the syntax would be
```rust
#[builder(setter(strip_option(fallback = outer_opt))]
```
Which I prefer. I use strings for prefix settings because they get concatenated into other identifiers (which has some implications - e.g. a prefix can be a reserved word) but if the identifier is used as is I prefer to keep it `syn::Ident` all the way.

typed-builder-macro/src/field_info.rs
```diff
@@ -345,6 +378,58 @@ impl ApplyMeta for SetterSettings {
...
+                result = Err(Error::new_spanned(
+                    arg.name(),
+                    format!("Too many paramters {:?}", arg.name().to_string()),
+                ));
+
+                continue;
+            }
+
+            if arg.name().to_string().as_str() != "fallback" {
+                result = Err(Error::new_spanned(
```
idanarye, line 414: Nit: as I've said before that this entire thing should be replaced with the `ApplyMeta` style, but just want to point out that instead of setting `result` here and `continue`-ing, you could have just `return`ed the error from here.

This also applies to the previous `if` in this loop, but more so to this one because it creates a bug. Consider:
```rust
#[builder(setter(strip_option(invalid = setting, fallback = "outer_opt"))]
```

The first iteration will set `result` to `Err` for the `invalid = setting` - but then the next iteration will set it to `Ok` for the `fallback`.

idanarye (changes requested):
I have some remarks on this. Do you want to fix them? If not, I don't mind doing them myself after merging.

typed-builder-macro/src/field_info.rs
ifiokjr: Nice, I've applied this.

typed-builder-macro/src/field_info.rs
ifiokjr: Totally agree, and updated.

typed-builder-macro/src/field_info.rs
ifiokjr: I've updated this, but the only problem is that the code now allows for:

- multiple sub attributes with the same name `strip_option(fallback = a, fallback = b)` 
- empty sub attributes `strip_option()`

If I update the `apply_sub_attr` in the `ApplyMeta` trait it might be a breaking change for code bases that may rely on this behaviour. I could override locally I suppose.

Is this important or should I ignore for now?

ifiokjr:
@idanarye should I create a separate PR for `strip_bool`?

idanarye:
Yes, but let's merge this one first. Much of it would probably be the same so lets iron it out once instead of working on the same issues twice.

typed-builder-macro/src/field_info.rs
```diff
@@ -345,6 +378,34 @@ impl ApplyMeta for SetterSettings {
...
+        Self { fallback: None, span }
+    }
+}
+
+impl ApplyMeta for Strip {
+    fn apply_meta(&mut self, expr: AttrArg) -> Result<(), Error> {
+        match expr.name().to_string().as_str() {
+            "fallback" => {
+                let ident: syn::Ident = expr.key_value().map(|kv| kv.parse_value())??;
+                self.fallback = Some(ident);
```
idanarye, line 398: Since [you were concerened](https://github.com/idanarye/rust-typed-builder/pull/150#discussion_r1727107584) about "multiple sub attributes with the same name" - this is the place to handle this. If we were using `apply_flag_to_field` here, it would have checked internally that `self.fallback.is_none()` and if not it would have thrown an error. You can:

1. Do the same here (without the helper) - just throw an error if `self.fallback.is_none()`.
2. Create a new helper method, similar to `apply_flag_to_field`, that does it internally.

I think the second option can get a bit complicated, and I'm not sure it's worth doing RN.

About your other concern - "empty sub attributes" - I don't consider that an issue.

typed-builder-macro/src/field_info.rs
```diff
@@ -335,7 +335,40 @@ impl ApplyMeta for SetterSettings {
...
             "skip" => expr.apply_flag_to_field(&mut self.skip, "skipped"),
             "into" => expr.apply_flag_to_field(&mut self.auto_into, "calling into() on the argument"),
-            "strip_option" => expr.apply_flag_to_field(&mut self.strip_option, "putting the argument in Some(...)"),
+            "strip_option" => {
+                let caption = "putting the argument in Some(...)";
+
+                match expr {
+                    AttrArg::Sub(sub) => {
+                        let span = sub.span();
+                        let mut strip_option = Strip::new(span);
```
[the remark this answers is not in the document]
ifiokjr: Fixed

ifiokjr:
> Yes, but let's merge this one first. Much of it would probably be the same so lets iron it out once instead of working on the same issues twice.

I'll create a follow up PR once this is merged.

PR #151, opened 2024-08-22
feat: add `strip_option(fallback = field_bool)`

ifiokjr:
### Description

Follow up from #150.


ifiokjr:
@idanarye this is ready for review

PR #159, opened 2025-03-11
Enhance TypedBuilder: Support Struct-Level Defaults for Fields

Tockra:
#### **Summary**  
This PR enhances the `TypedBuilder` derive macro by allowing users to define field defaults at the **struct level** instead of tagging each field individually. This simplifies usage and reduces boilerplate in structs with many optional fields.

#### **New Feature**  
- Introduced a **struct-level `#[builder(field_global_defaults(...))]` attribute**, which applies settings to all fields unless explicitly overridden.
- Fields now **inherit global defaults** unless they have specific attributes.

#### **Changes Implemented**  
1. **Extended `TypeBuilderAttr`**
   - Added `field_global_defaults` to store struct-level field settings.
   - Modified `ApplyMeta` to recognize `field_global_defaults` and parse settings.

2. **Updated `FieldInfo`**
   - Fields now merge their own attributes with global defaults.
   - Fields explicitly specifying attributes will **override global defaults**.

3. **Refactored `StructInfo`**
   - Passes struct-level defaults to `FieldInfo` when creating field definitions.

#### **Example Usage**
**Before (Without Struct-Level Defaults):**
```rust
#[derive(TypedBuilder)]
struct Foo {
    #[builder(default, setter(strip_option))]
    x: Option<i32>,   
    #[builder(default, setter(strip_option))]
    y: Option<i32>,
}
```

**After (With Struct-Level Defaults):**
```rust
#[derive(TypedBuilder)]
#[builder(field_global_defaults(default, setter(strip_option)))]
struct Foo {
    x: Option<i32>,  // Inherits `default` and `strip_option`
    y: Option<i32>,  // Inherits `default` and `strip_option`
}
```

#### **Why This Change?**
- **Reduces repetitive annotations** for structs with many optional fields.
- **Maintains backward compatibility**—existing field-specific annotations continue to work.

#### **Not included**
- allow only to overwrite Option<> related fields while non Option related fields stays untouched
- allow to overwrite the global setting on field level

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -991,3 +991,19 @@ fn test_mutators_for_generic_fields() {
 
     assert_eq!(Foo::builder().x_plus(1).y(2).build(), Foo { x: 1, y: 2 });
 }
+
+#[test]
+fn test_fields_global_defaults() {
+    #[derive(Debug, PartialEq, TypedBuilder)]
+    #[builder(field_global_defaults(default, setter(strip_option)))]
+    struct Foo {
+        x: Option<u32>,
+        y: Option<i32>,
+        z: Option<i32>,
+    }
+
+    let foo = Foo::builder().x(2).z(3).build();
+    assert_eq!(foo, Foo { x: Some(2), y: None, z: Some(3) });
+    let foo = Foo::builder().x(5).y(4).z(2).build();
+    assert_eq!(foo, Foo { x: Some(5), y: Some(4), z: Some(2) });
+}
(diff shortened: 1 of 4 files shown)


Tockra:
> This is already supported with `field_defaults`:
> 
> ```rust
> #[derive(TypedBuilder)]
> #[builder(field_defaults(default, setter(strip_option)))]
> struct Foo {
>     x: Option<i32>,   
>     y: Option<i32>,
> }
> ```

Damn, thank you for the information ;)

PR #160, opened 2025-03-19
Improve `field_defaults(strip_option)` semantics

### LINKED ISSUE #153: Improve `field_defaults` semantics
### Description

Here is the current documentation for the `field_defaults(...)` attribute on `TypedBuilder` structs.

> is structured like the #[builder(...)] attribute you can put on the fields and sets default options for fields of the type. If specific field need to revert some options to the default defaults they can prepend ! to the option they need to revert, and it would ignore the field defaults for that option in that field.

I have the following proposal. 

1. For the attributes `strip_option` and `strip_bool` there should be a way to only apply them to fields that are valid. Currently, when setting `strip_option` you need to use `!` to negate it on non-option fields. This should not be required. Perhaps `ignore_invalid` as a sub attribute flag to prevent this being a breaking change. 
2. In #150, support was added for fallback methods with this syntax `strip_option(fallback = custom_name_opt)`. This currently doesn't work for `field_defaults`. There should be an option available for `strip_option` which allows setting a `fallback_prefix` and `fallback_suffix` which adds the fallback to all optional methods automatically. 

### Possible API

```rust
use typed_builder::TypedBuilder;

#[derive(TypedBuilder)]
#[builder(field_defaults(default, setter(strip_option(ignore_invalid, fallback_suffix = "_opt")))]
struct Foo {
    // Defaults to None, options-stripping is performed and a fallback 
    // method of `x_opt` is available.
    x: Option<i32>,

    // Defaults to 0, no longer need `!strip_option`:
    y: i32,

    // Defaults to Some(13), option-stripping is performed and a fallback 
    // method of `z_opt` is available:
    #[builder(default = Some(13))]
    z: Option<i32>,
}
```

ifiokjr:
### Description

Fixes #153

- Added `ignore_invalid`, `fallback_prefix`, and `fallback_suffix` options to `strip_option` for better customization.
- Improved handling of `strip_option` to support field defaults and non-Option fields when `ignore_invalid` is enabled.
- Added tests for new `strip_option` features to ensure functionality.

### DIFF
--- a/typed-builder-macro/src/struct_info.rs
+++ b/typed-builder-macro/src/struct_info.rs
@@ -261,8 +261,10 @@ impl<'a> StructInfo<'a> {
 
         let deprecated = &field.builder_attr.deprecated;
 
+        let options_was_stripped;
         let arg_type = if field.builder_attr.setter.strip_option.is_some() && field.builder_attr.setter.transform.is_none() {
             if let Some(inner_type) = field.type_from_inside_option() {
+                options_was_stripped = true;
                 inner_type
             } else if field
                 .builder_attr
@@ -271,6 +273,7 @@ impl<'a> StructInfo<'a> {
                 .as_ref()
                 .map_or(false, |s| s.ignore_invalid)
             {
+                options_was_stripped = false;
                 field_type
             } else {
                 return Err(Error::new_spanned(
@@ -279,6 +282,7 @@ impl<'a> StructInfo<'a> {
                 ));
             }
         } else {
+            options_was_stripped = false;
             field_type
         };
         let (arg_type, arg_expr) = if field.builder_attr.setter.auto_into.is_some() {
@@ -298,17 +302,13 @@ impl<'a> StructInfo<'a> {
         let strip_option_fallback = field.builder_attr.setter.strip_option.as_ref().and_then(|strip_option| {
             if let Some(ref fallback) = strip_option.fallback {
                 Some((fallback.clone(), quote!(#field_name: #field_type), quote!(#arg_expr)))
-            } else if let (Some(prefix), Some(suffix)) = (&strip_option.fallback_prefix, &strip_option.fallback_suffix) {
+            } else if strip_option.fallback_prefix.is_none() && strip_option.fallback_suffix.is_none() {
+                None
+            } else {
+                let prefix = strip_option.fallback_prefix.as_deref().unwrap_or_default();
+                let suffix = strip_option.fallback_suffix.as_deref().unwrap_or_default();
                 let fallback_name = syn::Ident::new(&format!("{}{}{}", prefix, field_name, suffix), field_name.span());
                 Some((fallback_name, quote!(#field_name: #field_type), quote!(#arg_expr)))
-            } else if let Some(prefix) = &strip_option.fallback_prefix {
-                let fallback_name = syn::Ident::new(&format!("{}{}", prefix, field_name), field_name.span());
-                Some((fallback_name, quote!(#field_name: #field_type), quote!(#arg_expr)))
-            } else if let Some(suffix) = &strip_option.fallback_suffix {
-                let fallback_name = syn::Ident::new(&format!("{}{}", field_name, suffix), field_name.span());
-                Some((fallback_name, quote!(#field_name: #field_type), quote!(#arg_expr)))
-            } else {
-                None
             }
         });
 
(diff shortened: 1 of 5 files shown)


typed-builder-macro/src/struct_info.rs
```diff
@@ -283,13 +294,23 @@ impl<'a> StructInfo<'a> {
...
-            .setter
-            .strip_option
-            .as_ref()
-            .and_then(|strip_option| strip_option.fallback.as_ref())
-            .map(|fallback| (fallback.clone(), quote!(#field_name: #field_type), quote!(#arg_expr)));
+
+        let strip_option_fallback = field.builder_attr.setter.strip_option.as_ref().and_then(|strip_option| {
+            if let Some(ref fallback) = strip_option.fallback {
+                Some((fallback.clone(), quote!(#field_name: #field_type), quote!(#arg_expr)))
+            } else if let (Some(prefix), Some(suffix)) = (&strip_option.fallback_prefix, &strip_option.fallback_suffix) {
```
idanarye, line 301: Instead of handling all these cases, how about (I didn't try to compile it, so some changes may be needed to get it to typecheck):

```rust
} else if strip_option.fallback_prefix.is_none() && strip_option.fallback_suffix.is_none() {
    None
} else {
    let prefix = strip_option.fallback_prefix.as_deref().unwrap_or_default();
    let suffix = strip_option.fallback_suffix.as_deref().unwrap_or_default();
    let fallback_name = syn::Ident::new(&format!("{}{}{}", prefix, field_name, suffix), field_name.span());
    Some((fallback_name, quote!(#field_name: #field_type), quote!(#arg_expr)))
}
```

typed-builder-macro/src/struct_info.rs
```diff
@@ -298,7 +319,11 @@ impl<'a> StructInfo<'a> {
             let body = &transform.body;
             (quote!(#(#params),*), quote!({ #body }))
         } else if field.builder_attr.setter.strip_option.is_some() {
-            (quote!(#field_name: #arg_type), quote!(Some(#arg_expr)))
+            if field.type_from_inside_option().is_some() {
```
idanarye, line 322: Instead of calling `type_from_inside_option` twice, how about having a flag indicating what happened the first time you called it?

```rust
let option_was_stripped;
let arg_type = if field.builder_attr.setter.strip_option.is_some() && field.builder_attr.setter.transform.is_none() {
    if let Some(inner_type) = field.type_from_inside_option() {
        option_was_stripped = true;
        inner_type
    } else if field
        .builder_attr
            .setter
            .strip_option
            .as_ref()
            .map_or(false, |s| s.ignore_invalid)
    {
        option_was_stripped = false;
        field_type
    } else {
        return Err(Error::new_spanned(
                field_type,
                "can't `strip_option` - field is not `Option<...>`",
        ));
    }
} else {
    option_was_stripped = false;
    field_type
};
```

With this, instead of three cases here (`strip_option`, `strip_option` on non-`Option` that was ignored, no `strip_option`) you just need to handle two cases (`option_was_stripped` being `true` or `false`)

ifiokjr: Makes a lot more sense!

typed-builder-macro/src/struct_info.rs
```diff
@@ -261,13 +261,28 @@ impl<'a> StructInfo<'a> {
 
         let deprecated = &field.builder_attr.deprecated;
 
-        // NOTE: both auto_into and strip_option affect `arg_type` and `arg_expr`, but the order of
-        // nesting is different, so we have to do this little dance.
+        let options_was_stripped;
```
idanarye, line 264: Nit: should be "option" and not "options".

typed-builder-macro/src/struct_info.rs
```diff
@@ -298,7 +319,11 @@ impl<'a> StructInfo<'a> {
             let body = &transform.body;
             (quote!(#(#params),*), quote!({ #body }))
         } else if field.builder_attr.setter.strip_option.is_some() {
-            (quote!(#field_name: #arg_type), quote!(Some(#arg_expr)))
+            if options_was_stripped {
```
idanarye, line 322: Nit: we can check this variable instead of the `field.builder_attr.setter.strip_option.is_some()` in the line above. If `strip_option` is `None`, the flag will be false anyway and we won't wrap `arg_expr` in a `Some`.

idanarye (approved):
I have some small nitpicks, but I'll just fix them myself after merging.

